PR #303, opened 2014-03-04
Add data import of data validations to reader

alsp80:
This should add import of data validations in a worksheet.


alsp80:
Seems I've worked on an outdated copy. I will try to send a pull request with an updated source code version. New to GitHub, sorry for any inconvenience caused.

PR #310, opened 2014-03-14
More sophisticated calculation of cell widths

chrassig:
Let me start by saying we really love the ABAP2XLSX project. Compared to exporting CSV files and importing them to Excel, it makes transferring data from SAP systems to Excel and vice versa so much easier.

What we did not like so far was the optimization of column width, especially because of the way this is realized in the XLSX file format (why did they not just add an attribute "optimize_column_width=true"?). Anyway, ABAP2XLSX being an open source project, we decided to try to bring some progress to this functionality.

We ended up rewriting the method CALCULATE_CELL_WIDTH of the class ZCL_EXCEL_WORKSHEET.

We added the following features:
a) Before, the calculation of the cell width in ABAP2XLSX did not take into account the fact that Excel uses a proportional font by default. Thus, columns with many "i"s ended up too wide, while columns containing many "W"s ended up not wide enough.

We now use the function module LOAD_FONT, contained in SAP package BC. Thus, it should be available in pretty much every SAP system that runs ABAP. LOAD_FONT returns an internal table containing the width of every letter known in a font. You can upload Excel's default font Calibri from the regular .TTF font files using transaction SM73. Make sure to upload all four files (standard, bold, italic, bold and italic) and to use the description "Calibri", i.e. exactly the name Excel displays for the font. If you want to use other fonts in your files created by ABAP2XLSX, you can of course upload other fonts as well.

We calculate the sum of the widths of all characters in a cell, and add some cell padding, similar to the way Excel does it. If the font is not available in your SAP system, we revert back to a logic similar to the old one.

As calling LOAD_FONT for every cell took several seconds when we tested with larger tables, we store the values read from the function in a hash table. Thus, we only have to call the function once for every combination of font name and font attributes used in an Excel file.

b) XLSX stores dates in a cryptic format. For dates, the current implementation calculates the width of the cryptic values, thus, the columns end up too narrow.

Fortunately, ABAP2XLSX already contains a method to transform the cryptic value into an ABAP date. We then use ABAP's WRITE statement to create a date in the format the user is preferring, and calculate the width of this formatted date.

c) ABAP2XLSX currently does not consider auto filters. Thus, a column may end up too narrow, not showing the complete header text.

We fixed this by adding a fixed value to the cell width if an auto filter is activated. We use a fixed value because the size of the auto filter button in Excel does not depend on the font size you use.

d) ABAP2XLSX only estimates the width of the characters. Excel adds a small fixed amount to the width of the characters to create some horizontal cell padding. We added cell padding as well, as this brought the cell width even closer to the values calculated by Excel.

e) ABAP2XLSX already uses ABAP's data type FLOAT to store the width of a cell. However, CALCULATE_CELL_WIDTH used data type I (integer) to calculate the cell width before. Changing the parameter type for the parameter EP_WIDTH from I to FLOAT, and adjusting the variable used by the calling method CALCULATE_COLUMN_WIDTHS in the same way, made the calculation even more precise.

### DIFF
--- a/ZA2X/CLAS/ZCL_EXCEL_WORKSHEET.slnk
+++ b/ZA2X/CLAS/ZCL_EXCEL_WORKSHEET.slnk
@@ -1,5 +1,26 @@
 <?xml version="1.0" encoding="utf-8"?>
 <CLAS CLSNAME="ZCL_EXCEL_WORKSHEET" VERSION="1" LANGU="E" DESCRIPT="Worksheet" CATEGORY="00" EXPOSURE="2" STATE="1" RELEASE="0" CLSFINAL="X" CLSCCINCL="X" FIXPT="X" UNICODE="X" CLSBCCAT="00" DURATION_TYPE="0 " RISK_LEVEL="0 ">
+<types CLSNAME="ZCL_EXCEL_WORKSHEET" CMPNAME="MTY_S_FONT_METRIC" VERSION="1" LANGU="D" EXPOSURE="0" STATE="1" EDITORDER="1 " TYPTYPE="4" SRCROW1="6 " SRCCOLUMN1="4 " SRCROW2="9 " SRCCOLUMN2="32 " TYPESRC_LENG="143 " TYPESRC="BEGIN OF mty_s_font_metric,
+           char       TYPE c LENGTH 1,
+           char_width TYPE tdcwidths,
+         END OF mty_s_font_metric
+"/>
+ <types CLSNAME="ZCL_EXCEL_WORKSHEET" CMPNAME="MTY_TH_FONT_METRICS" VERSION="1" LANGU="D" EXPOSURE="0" STATE="1" EDITORDER="2 " TYPTYPE="4" SRCROW1="11 " SRCCOLUMN1="4 " SRCROW2="13 " SRCCOLUMN2="30 " TYPESRC_LENG="105 " TYPESRC="mty_th_font_metrics
+           TYPE HASHED TABLE OF mty_s_font_metric
+           WITH UNIQUE KEY char
+"/>
+ <types CLSNAME="ZCL_EXCEL_WORKSHEET" CMPNAME="MTY_S_FONT_CACHE" VERSION="1" LANGU="D" EXPOSURE="0" STATE="1" EDITORDER="3 " TYPTYPE="4" SRCROW1="15 " SRCCOLUMN1="4 " SRCROW2="21 " SRCCOLUMN2="31 " TYPESRC_LENG="306 " TYPESRC="BEGIN OF mty_s_font_cache,
+           font_name       TYPE zexcel_style_font_name,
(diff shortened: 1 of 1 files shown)


ivanfemia:
thank you for this great improvement I will test it locally and include it in the next daily build.

chrassig:
Hi Ivan,

thanks a lot for your incredibly fast response. In this way, contributing really is fun :)

Best regards,
Christian

ivanfemia:
Christian,
I tried the change in my system, but unfortunately even if I imported Calibri in the system report DEMO31 it doesn't get the changes.
I found that this is caused by the FM LOAD_FONT that doesn't return any metric. (caused by FM GENERATE_FONT, the function SAPSCRIPT_FONTS_GET_SINGLE returns exception NOT_FOUND - "ERROR, printer font doesn't exist")
Do I have to perform any special activity after the import of the font from SE73?

chrassig:
I am not sure. You might try to run the report RSPOCFTEST1. Select SWIN (or PDF1) as "device type", and the 8-letter all upper-case "font name" you chose when you uploaded the font (ZCALIBRI in our system, I suggest you use the value help).

Execute the report, and you should see a table with the character widths of a sample string.

We have executed this report many types while testing our changes. In its source code, I saw that it also generates data in the database. This data might be needed to run LOAD_FONT successfully.

I guess we both agree that we should write a tutorial on how to use SE73 and what to do afterwards to make LOAD_FONT work.

ivanfemia:
Report works fine...
I get the metrics. I will try to investigate further on this.
A little guide on SE73 would be great (even if it is pretty simple)

chrassig:
Did you try to run DEMO31 again after you ran RSPOCFTEST1?

ivanfemia:
It worked with a re-import of the font...

alsp80:
Excellent addition, loving it. One remark regarding upload of Calibri: I had to change the name after I've uploaded all four TTF files because SE73 will add 'NBIX' to the name of the font in order to mark which font types are available. As soon as I had changed the name back to 'Calibri' everything was fine and cell widths are now calculated correctly.

I have one further request on this (which I will also look into implementing myself): in case rowheight and setup of the cell (wrap text) allow text wrapping I would like to calculate the cell width accordingly.

Anyway, thanks for the contribution and the fast pull.

chrassig:
Sorry, I was on holiday for the previous two weeks.

alsp80, you are right. We also had to rename the font manually after importing it. This is one point that should be covered in the tutorial / little guide we talked about before.

Word wrapping is one of many features that the method does not consider at the moment. Other features of XLSX that we do not consider yet include different formats (e.g. times line 11:30:00), the results of formulas (currently only fixed values are considered, cells that contain formulas are treated like empty cells), auto-filters that are manually defined (not part of a table) and probably many more. Our focus was to enable an automated export from an ALV grid to an XLSX file. For other use cases, different features like the one you have mentioned definitely make sense.

PR #319, opened 2014-05-02
Fixed a bug in shared string table reading

rplantiko:
There was a bug in reading strings from the SST, caused by not catching the "node value" event in the main loop. Shared strings could point wrongly to the first table element (which usually contains the empty string). My tests didn't detect the bug since I worked with tables containing only one string. I extended them to catch this bug, before fixing it.
The delta display is a bit weird, probably due to different line feed symbols? My changes are mainly in method READ_WORKSHEET_DATA


gregorwolf:
Dear Rüdiger,

please make sure you have the latest SAPlink version and also set your pretty printer to keyword uppercase.

Best regards
Gregor

PR #333, opened 2014-10-02
Ignore merged cells when calculating column width

oliver-huetkoeper:
When calculating column width cells that are merged should be ignored. This is MS Excel behavior and should be handled similar in ABAP2XLSX.


gregorwolf:
Hi Oliver,

sorry for the late response to your pull request. Could you please do the following:
- add a Unit Test for the Method IS_CELL_IN_RANGE in class ZCL_EXCEL_COMMON
- Enhance the demo report zdemo_excel31 to contain also an example for a merged cell so the behaviour is also demonstrated in a test output

Best regards
Gregor

oliver-huetkoeper:
Hi Gregor,

sure, I will add this.

Best regards,
Oliver

oliver-huetkoeper:
Hi Gregor,

I have added the test method and enhanced the demo.

Best regards,
Oliver

PR #351, opened 2015-01-08
Patch with all slinkees exported by ZAKE

mvcube:
Hi Ivan,

unfortunately, all the slinkees are different from their originals after export with ZAKE. You will have to compare the differences for yourself.

Marcus


review-ninja:
[ReviewNinja](http://app.review.ninja/ivanfemia/abap2xlsx/pull/351)

gregorwolf:
Hi Marcus,

as all files are affected I would think that it is a problem with the Line Break. Can you check with a editor like Notepad++ what's the linebreak of the files in the original repository and in your export?

Best regards
Gregor

mvcube:
Hi Gregor,

the line termination is indeed different but I assume that GIT handles this kind of difference automatically for text files. I have the feeling that the slinkees in the repository do not completely represent the current stat of development (what is in the nugget). As an example compare the interface ZIF_EXCEL_READER which seems to have been enhanced since the last download of its slink file. I didn't touch it so I'm sure I'm not responsible for the differences.

In order to check what has been actually changed by me, Tortoise SVN's visual diff ("compare to previous version") can be of help.

ivanfemia:
Hi Marcus,

were you be able to improve the exported files?
Your enhancement is really interesting and I would like to include it in the main stream!

Thank you
Ivan

mvcube:
How should I? I've just used ZAKE for the export. What should I have done differently? I have only limited access to the customer's system so my chances to do more than I've done until now are very limited.

I suggest you just load my version into your system and try it out. I can't image that I have seriously broken anything.

StefanSchmoecker:
Marcus,

when you say "Nugget" - did you install ABAP2XLSX from the "latest build" nugget or individually from the SVN-directory?
If you did install from the latest-build nugget:  Did you update the objects you touched in your system from the repository before adding your changes and using ZAKE?

mvcube:
Here is what I did:
I downloaded the "latest build" and installed it to the SAP system with SAPLink. This was some months ago. Then I did a bunch of modifications to the code.

Only recently I discovered ZAKE and used it to export the individual slinkees to the notebook provided by my customer. (I cannot access GITHub from this PC.) I then used an external USB storage device to move the slinkees to my private system where I had checked out the repository.

There are several possible reasons for the differences between my export and the original slnk files:
- The customer's SAP system is a 7.3 Netweaver (BW). I don't know the release status of Ivan's system.
- The base operating system is Unix, not Windows.
- The package name of my build is ZEXCEL, not ZA2X.
- My user id is different from Ivan's.
- ZAKE might have a different release level then the version used by Ivan.
- The slinkees in the repository might not be the same code as the nugget I downloaded.

I've seen major differences in objects that I definitely did not touch (ZIF_EXCEL_READER). I've no idea what I can do better.

Marcus

mvcube:
I had a nice conversation with Stefan Schmöcker. He will take care of the modifications. I therefore close this request.

Issue #323, opened 2014-06-17
ZCL_EXCEL_CONVERTER->EXECUTE_CONVERTER error when biding grid


fidley:
Hello, 
when we have a subclass of cl_gui_alv_grid and we want to bind it then converter doesn't bind the object as it's not of one of allowed kind. In my installation I've corrected the method from title to check also if a superclass in one of the allowed and if yes then I continue to proces converting. Here is full code of changed method. I'm not sure if this is not done already in newest builds.

Regards
Łukasz

method execute_converter.
  data: lt_fieldcatalog         type zexcel_t_fieldcatalog,
        ls_fieldcatalog         type zexcel_s_converter_fcat,
        lo_if                   type ref to zif_excel_converter,
        ls_types                type ts_alv_types,
        lo_addit                type ref to cl_abap_classdescr.
  data: lo_addit_superclass     type ref to cl_abap_classdescr.

  if io_object is bound.
    try.
        lo_addit            ?= cl_abap_typedescr=>describe_by_object_ref( io_object ).
      catch cx_sy_move_cast_error.
        raise exception type zcx_excel.
    endtry.
    ls_types-seoclass = lo_addit->get_relative_name( ).
    read table wt_objects into ls_types with table key seoclass = ls_types-seoclass.
    if sy-subrc ne 0.
      do.
        free lo_addit_superclass.
        lo_addit_superclass  = lo_addit->get_super_class_type( ).
        if lo_addit_superclass is initial.
          sy-subrc = '4'.
          exit.
        endif.
        lo_addit = lo_addit_superclass.
        ls_types-seoclass = lo_addit->get_relative_name( ).
        read table wt_objects into ls_types with table key seoclass = ls_types-seoclass.
        if sy-subrc eq 0.
          exit.
        endif.
      enddo.
    endif.

```
if sy-subrc = 0.
  create object lo_if type (ls_types-clsname).

  try.
    lo_if->create_fieldcatalog(
      exporting
        io_object       = io_object
        it_table        = it_table
      importing
        es_layout       = ws_layout
        et_fieldcatalog = wt_fieldcatalog
        eo_table        = wo_table
        et_colors       = wt_colors
        et_filter       = wt_filter
           ).
  endtry.
```
- data lines of highest level.
    if ws_layout-max_subtotal_level > 0.
      add 1 to ws_layout-max_subtotal_level.
    endif.
  else.
    raise exception type zcx_excel.
  endif.
  else.
[...]

PR #352, opened 2015-01-10
#323 excel converter

fidley:
Allowed to used objects which inherits from cl_gui_alv_grid


review-ninja:
[ReviewNinja](http://app.review.ninja/ivanfemia/abap2xlsx/pull/352)

ivanfemia:
Thank you Fidley, can you please also provide an example of usage... You can also create a new demo report

Ivan

fidley:
I've done some programs in which I created a class which is inheriting from cl_gui_alv_grid. In such classes I usually add some common handling of toolbar and other events. Once done and can be reused within the project, but when I wanted to use A2X to bind ALV grid from my class this was not possible because you read only name of the class of object which was passed to bind_alv method. So I've added a check if passed class is not inheriting from other and if yes it's checking if this is one from handled by A2X. That's all. I don't think it needs a demo report. It's rather bug correction.

ivanfemia:
Ok thank you for the explanation

PR #375, opened 2015-05-18
Style of remaining columns with no values

oliver-huetkoeper:
When setting a style for all columns the reader only reads this style for filled columns. The style of the remaining columns is dropped.
Example: You activate border lines for all cells, but you fill only the first three columns. After reading and writing it with ABAP2XLSX, only the first three columns have border lines.

This is fixed by setting the default style here.


review-ninja:
<a href="http://app.review.ninja/ivanfemia/abap2xlsx/pull/375" target="_blank">ReviewNinja

ivanfemia:
Thank you, merging into the master...
I will include ASAP in the new daily build.

Issue #252, opened 2013-12-27
Error while using grid object and reuse_alv table.


ivanfemia:
When we use alv object taken from reuse_alv_grid_display and we use color column in source table then due to fact that color field has different length than in OO ALV then in zcl_excel_converter_alv->get_color we got an error.Â  This can be corrected with following code :
https://cw.sdn.sap.com/cw/docs/DOC52403

ivanfemia:
page is changing the link of the document. There should be "-1" between DOC and 52403

ivanfemia:
p882987,

is this still an open issue with the current daily build of ABAP2XLSX?  
I am also not able to read the link you supplied ( insufficient rights ) - could you please upload your suggestion as an attachment to this thread?

fidley:
Sorry that I come back so late but I totally forgot about this issue as I have corrected it directly on my instalation. I have no idea if in the new build it's still there or not cause I'm haven't install new version so far but here is the code I used:

method get_color.
  data: ls_con_col type zexcel_s_converter_col,
        ls_color   type ts_col_converter,
        l_line     type i,
        l_color(4) type c.
  field-symbols: <fs_tab> type standard table,
                 <fs_stab> type any,
                 <fs>      type any,
                 <fs_tcol>  type lvc_t_scol,
                 <fs_scol>  type lvc_s_scol,
                 <ft_slis> type standard table,
                 <fs_slis> type any.
- Loop trough the table to set the color properties of each line. The color properties field is
- Char 4 and the characters is set as follows:
- Char 1 = C = This is a color property
- Char 2 = 6 = Color code (1 - 7)
- Char 3 = Intensified on/of = 1 = on
- Char 4 = Inverse display = 0 = of
  
  assign io_table->\* to <fs_tab>.
  
  if ws_layo-info_fname is not initial or
     ws_layo-ctab_fname is not initial.
    loop at <fs_tab> assigning <fs_stab>.
      l_line = sy-tabix.
      if ws_layo-info_fname is not initial.
        assign component ws_layo-info_fname of structure <fs_stab> to <fs>.
        if sy-subrc = 0.
          if <fs> is not initial.
            l_color = <fs>.
            if l_color(1) = 'C'.
              read table wt_colors into ls_color with table key col = l_color+1(1)
                                                                int = l_color+2(1)
                                                                inv = l_color+3(1).
              if sy-subrc = 0.
                ls_con_col-rownumber  = l_line.
[...]

PR #399, opened 2015-08-28
#252 Error while using grid object and reuse_alv table.

fidley:
When you take the grid object from reuse_alv_grid_display FM then when
cells were colored I received dump as the structure of color table was
hardcoded to lvc_t_scol. When we have object from SLIS then there
structure is differen so I prepared code to handle both situations.


ivanfemia:
Thank you!

PR #472, opened 2017-01-24
Optimization

FoxAhead:
I was benchmarking exporting large tables to excel and found that big time take calls of methods:
ZCL_EXCEL_COMMON=>CONVERT_COLUMN2INT
ZCL_EXCEL_COMMON=>CONVERT_COLUMN2ALPHA
The most time take dealing with Unicode Code Point converting characters in and out (CL_ABAP_CONV*). 
image
This can be improved because when going for example in one column from top to bottom (as method BIND_TABLE do) you always have the same column coordinate and it is not necessary to calculate its address every time. 
image

So I was thinking about some cache table. But when you have too much iterations then even reading hashed table could take time in gross. But this was not tested.
The way I choose was to cache last value in methods CONVERT_COLUMN* in statics variables. This could give advantage when going column by column.
I've conducted some tests. Here is the test program.
```ABAP
*&---------------------------------------------------------------------*
*& Report  ZTEST_A2X
*&---------------------------------------------------------------------*
REPORT  ztest_a2x.
CONSTANTS: gc_save_file_name TYPE string VALUE 'ZTEST_A2X.xlsx'.
PARAMETERS:
  p_size TYPE i DEFAULT 10000.
START-OF-SELECTION.
  PERFORM main.
*&---------------------------------------------------------------------*
*&      Form  main
*&---------------------------------------------------------------------*
FORM main.
  DATA:
    lo_excel     TYPE REF TO zcl_excel,
    lo_worksheet TYPE REF TO zcl_excel_worksheet,
    lt_data      TYPE TABLE OF anla.
*--------------------------------------------------------------------*
  SELECT * INTO TABLE lt_data  FROM anla UP TO p_size ROWS.
  WHILE lines( lt_data ) BETWEEN 1 AND p_size.
    APPEND LINES OF lt_data TO lt_data.
  ENDWHILE.
  DELETE lt_data FROM p_size + 1.
  "Have exactly p_size records in lt_data[] here
  CREATE OBJECT lo_excel.
  lo_worksheet = lo_excel->get_active_worksheet( ).
  SET RUN TIME ANALYZER ON.
  lo_worksheet->bind_table( ip_table = lt_data ).
  lcl_output=>output(
    cl_excel            = lo_excel
    iv_writerclass_name = 'ZCL_EXCEL_WRITER_HUGE_FILE'
  ).
  SET RUN TIME ANALYZER OFF.
ENDFORM.                    "main
```
Two steps were measured: BIND_TABLE and OUTPUT. The first one deals with column address conversions. The second one deals with XML and this is not the subject of optimization - just for comparison.
Results are promising:
image
image
Method BIND_TABLE runs about 30% faster.
The time and number of calls to CL_ABAP_CONV* greatly decreased. It is not even on the first page.
image
Here it is filtered
image

### DIFF
--- a/ZA2X/CLAS/ZCL_EXCEL_COMMON.slnk
+++ b/ZA2X/CLAS/ZCL_EXCEL_COMMON.slnk
@@ -1362,6 +1362,9 @@ ENDMETHOD.</source>
               lv_module                       TYPE int4,
               lv_column                       TYPE zexcel_cell_column.
 
+  STATICS:    sv_prev_in  LIKE lv_column,
+              sv_prev_out LIKE ep_column.
+
 * Propagate zcx_excel if error occurs           &quot; issue #155 - less restrictive typing for ip_column
   lv_column = convert_column2int( ip_column ).  &quot; issue #155 - less restrictive typing for ip_column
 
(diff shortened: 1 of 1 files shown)


ivanfemia:
Thank you Fox for this enhancement

PR #516, opened 2018-01-30
CSV writer: fixed EOL length to handle CRLF correctly (Issue #519)

Cril2000:
Fixed EOL length to handle CRLF correctly

#519


gregorwolf:
+1

Issue #527, opened 2018-03-16
Performance problem after updating the version


bilencekic:
Hi Everyone,
I updated abap2xlsx version via abapgit but now i have a huge performance issue. Nothing changed, i am still using BIND_TABLE method with 16K rows and 60 columns. 
Last time it was taking 20-35 seconds with huge file writer but now it is nearly 150 seconds.
I tried SET_TABLE and same issue.
Any advice ?

ivanfemia:
Can you try to do a trace with se30 to identify the bottleneck?

bilencekic:
![xlsx](https://user-images.githubusercontent.com/17006134/37523248-52f7454c-2961-11e8-9027-fbd7eddbe4ce.png)
above is the SAT results, somehow GET_ROW method is taking %91 of total time.
above result is for 10K rows, bind_table is finishing in 23seconds for 16K rows, but excel writer takes so much time. i already have saplink of previois version, worst case i will go back to old version. I am thinking if there is any parameters that i need to set to make it faster.

ivanfemia:
Thanks,
I remember a similar issue in the past, I trying to recall what was the final outcome.

bilencekic:
thank you :D please try to remember  i need to transport requests to test system very soon. now i noticed system is using object_collection_iteration method after version update do get next row. inside iteration method it is using standard table maybe this can cause performance issue i didin't get strange issue. 
will it work if i just replace excel writer methods from the old one but keep other classes up to date ?

bilencekic:
ok i think i found the issue;
first of all that iterator doesn't look necessary, we can store in a global hash table. Each time itearator is instantiated table keep moving from one variable to another variable. 
![iterator](https://user-images.githubusercontent.com/17006134/37550638-b506e6ce-29cb-11e8-8b2b-814c7532438e.png)

and second thing is, program keep doing same search in each line. lets say there is row already in index 10, now it is searching if there is value in index 11, everytime it starts from the beggining. 
So in row number 2000, it works 2000 times to check if there is any value. But if there was a global variable like last_value_cell, it can start directly from 2000 to check.
![row](https://user-images.githubusercontent.com/17006134/37550645-cf7543e8-29cb-11e8-8816-0e258c91d0a7.png)

alright here is the test results;
same number of rows used with the updated log**ic, now it is less then 2 sec** @ivanfemia 

[...]

PR #529, opened 2018-03-20
Performance improvement for GET_ROW method

bilencekic:
Issue #527 has been fixed after updating ZCL_EXCEL_ROWS class by removing object iterator and introducing hash table.  ZCL_EXCEL_WORKSHEET->GET_ROW method has been updated accordingly based on new structure to read directly from hash table.
Before : 
image

And after:
image

### DIFF
--- a/src/zcl_excel_rows.clas.abap
+++ b/src/zcl_excel_rows.clas.abap
@@ -3,45 +3,54 @@
 *----------------------------------------------------------------------*
 *
 *----------------------------------------------------------------------*
-class ZCL_EXCEL_ROWS definition
-  public
-  final
-  create public .
+CLASS ZCL_EXCEL_ROWS DEFINITION
+  PUBLIC
+  FINAL
+  CREATE PUBLIC .
+
 
 *"* public components of class ZCL_EXCEL_ROWS
 *"* do not include other source files here!!!
 *"* protected components of class ZABAP_EXCEL_WORKSHEETS
 *"* do not include other source files here!!!
-public section.
+  PUBLIC SECTION.
+    TYPES:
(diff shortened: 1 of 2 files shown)


ivanfemia:
Thanks for your contribution
Do we still need the method GET_ITERATOR?

I would also suggest to have the DT_ROWS as protected, so no other class can change the content of the internal table.

bilencekic:
@ivanfemia it is easier to add READ ONLY statement, in this case it will be public which is easy to access and can only be changed inside the class;

```
    DATA:
          DT_ROWS TYPE HASHED TABLE OF T_ROWS WITH UNIQUE KEY ROW_INDEX READ-ONLY.
```

Get iterator we don't need i think it only exists in worksheet class, it is safe to delete.
I can do above changes as well if you want.

ivanfemia:
That's great!
thank you!

EMSINSANE:
Hi.  Just ( 27.03.18 ) installed the Project. 
Class ZCL_EXCEL_WORKSHEET contains Syntax-Errors  because the deleted Method ZCL_EXCEL_ROWS->GET_ITERATOR() is used in     ZCL_EXCEL_WORKSHEET->GET_ROWS_ITERATOR().

So Long

Insane

ivanfemia:
@EMSINSANE can you try to remove ZCL_EXCEL_WORKSHEET->GET_ROWS_ITERATOR() it should not be used anywhere... I have no access to a NetWeaver server at the moment.
You can either commit the change to the repository or tell us that this fixed the issue.

Thank you
Ivan

larshp:
The syntax error:
image

GET_ROWS_ITERATOR() is used in the following places,
image

Issue #528, opened 2018-03-16
Performance issue using zcl_excel_writer_2007


MLDOliveira:
Hi.

I'm starting to use ABAP2XLSX and my first was great, but I'm trying to generate a new XLSX file with 300.000,00 lines and the process is going really slow.

I'm using the code from ZDEMO_EXCEL_OUTPUTOPT_INCL.

This is a summary that I'm using in my code.

CREATE OBJECT go_excel.
gs_table_settings-table_style       = zcl_excel_table=>builtinstyle_medium2.
gs_table_settings-show_row_stripes  = abap_true.
gs_table_settings-nofilters         = abap_true.

"After fisrt declarations I fill the internal table  and use bind method
go_worksheet = go_excel->get_active_worksheet( ).
go_worksheet->set_title( "TABLE1" ).
go_worksheet->bind_table( ip_table          = it_001_saida10_exc
                                           is_table_settings = gs_table_settings ).

"After that I basically use the code below. 
lcl_output=>output( go_excel ).

"I'm doing a test right now using the ZCL_EXCEL_WRITER_HUGE_FILE class with the code below instead of code above

lcl_output=>output( cl_excel            = go_excel
                      iv_writerclass_name = 'ZCL_EXCEL_WRITER_HUGE_FILE' ).

My question is, there is a better way to work with huge files? What is the right way to export big files with many tabs?

@ivanfemia @larshp  could you help me?

ivanfemia:
Hi,

big files can have performances issue, I usually consider a batch process
for those.
Btw the class ZCL_EXCEL_WRITER_HUGE_FILE is the one
​ ​
that I
​
use
​ ​
at the moment
​ ​
in case of big files.

If anyone has a better idea to improve this process, I'm available to
discuss.

Ivan

MLDOliveira:
@ivanfemia do you have an example of using the class ZCL_EXCEL_WRITER_HUGE_FILE?
I'm trying to use it but I'm not moving forward

ivanfemia:
Just use that class instead of the classical writer.
It has some limitation on the features available

MLDOliveira:
@ivanfemia the problem is when I use the code below

MLDOliveira:
@ivanfemia the problem is when I use the code below I get a erro message saing that "the method CREATE_XL_SHEET is protected"

CREATE OBJECT cl_writer_huge TYPE zcl_excel_writer_huge_file.
[...]

PR #544, opened 2018-07-14
Performance improvement for GET_ROW/GET_COLUMN

Pacojka:
Hallo Ivan,

i implemented the changes as described in issue #528 for rows and columns objects.

I kept the object_collection and the get_iterator method because lack of time and the visibility (public) of the get_iterator method.

In the worksheet class, I changed get_row and get_column.


Pacojka:
I tested the change using ZDEMO_EXCEL15. Everything looks fine. 
The performance improvement is as expected.

ivanfemia:
Thanks.
I think that we should improve this removing completely the iterator and the few methods no longer needed... But first of all we need to mark those as obsolete

PR #561, opened 2018-09-16
propragate exceptions

AtomKrieg:
add "RAISING ZCX_EXCEL" at end of some method's definitions


AtomKrieg:
btw. what do you think about exception syntax change (code bloat reduce)
```
zcx_excel=>raise_text( 'Some exception text'(001) ).
```

instead
```
raise exception type zcx_excel
  exporting
    error = 'Some exception text'(001).
```

ivanfemia:
I like the more compact code for exception raising.
Some code was used to keep compatibility to old releases, but I think that now this should not be a big issue,

Thank you for your contribution!

Issue #364, opened 2015-03-12
style at a merged cell 


nxdiez:
Let me say it is the first time I collaborate at GI, and I do not know how it works really well.
I hope this is the right place to write.
Secondly, I discovered ABAP2XLSX and it is awesome! Thank you, guys!

While using ABAP2XLSX, I have faced that applying an style to a merged cell, it just affects to top-left cell of the group of cells, Checkin the example ZDEMO_EXCEL13, I see I had to apply the style to all the cells, and set the value to the top-left cell.
In my case, I had many merge cells, so I decided to create a new method SET_MERGE_STYLE on ZCL_EXCEL_WORKSHEET and modify SET_MERGE.

Bellow the code,
1 - SET_MERGE_STYLE definition

``` abap
methods SET_MERGE_STYLE
    importing
      !IP_COLUMN_START type ZEXCEL_CELL_COLUMN_ALPHA optional
      !IP_COLUMN_END type ZEXCEL_CELL_COLUMN_ALPHA optional
      !IP_ROW type ZEXCEL_CELL_ROW optional
      !IP_ROW_TO type ZEXCEL_CELL_ROW optional
      !IP_STYLE type ZEXCEL_CELL_STYLE optional .
```

2 - SET_MERGE_STYLE implementation

``` abap
METHOD set_merge_style.
  DATA: ld_row_start TYPE zexcel_cell_row,
        ld_row_end TYPE zexcel_cell_row,
        ld_column_start TYPE zexcel_cell_column,
        ld_column_end TYPE zexcel_cell_column,
        ld_current_column TYPE zexcel_cell_column_alpha,
        ld_current_row type zexcel_cell_row.

  MOVE: ip_row_to TO ld_row_end,
        ip_row    TO ld_row_start.
  IF ld_row_end IS INITIAL.
    ld_row_end = ld_row_start.
  ENDIF.
  ld_column_start = zcl_excel_common=>convert_column2int( ip_column_start ).
  ld_column_end   = zcl_excel_common=>convert_column2int( ip_column_end ).
  IF ld_column_end IS INITIAL.
    ld_column_end = ld_column_start.
  ENDIF.
  "set the style cell by cell
  WHILE ld_column_start <= ld_column_end.
    ld_current_column = zcl_excel_common=>convert_column2alpha( ld_column_start ).
    ld_current_row = ld_row_start.
    WHILE ld_current_row <= ld_row_end.
[...]

PR #574, opened 2018-12-14
style at a merged cell #364

bohrh:
I have implemented the suggested changes and are using it at a customer project. It's working like a charm


ivanfemia:
Thank you @bohrh , can you provide also a sample report to add to our library that shows how to leverage on this functionality?

Thank you
Ivan

ivanfemia (approved):
Thanks

PR #577, opened 2019-01-09
Issue #572 - Protect sheet with filter caused Excel error

MCH123:
Issue #572 - Protect sheet with filter caused Excel error 

Autofilter + sheet protection causes excel error.

Sheet protection must be set BEFORE autofilter in XML.


ivanfemia:
Thank you for your contribution

PR #594, opened 2019-02-17
fix #587

### LINKED ISSUE #587: Comment not displayed in second sheet
comments are not displaying in the second sheet if no comments set for the first sheet. Sample program is attached. 

[zjk_excel_comment.abap.txt](https://github.com/ivanfemia/abap2xlsx/files/2827310/zjk_excel_comment.abap.txt)

AtomKrieg:
Issue are caused by associating file numbering with sheet number.
Files (like comments#.xml) should be numbered as they appear.


gregorwolf:
Can you please upgrade your abapGit installation so w don't get so many changed files.

AtomKrieg:
I accidentally compared the same versions of abapGit and consider that was using latest one.

PR #611, opened 2019-04-26
Option to create empty file or not

kwaishang:
Two small patches :
- Allow INT8 column types
- Add the optionnal parameter in bind table method to create empty file or not if the input internal table is empty.


gregorwolf:
Hi @kwaishang could you adjust or contribute a test report where this feature is used?

kwaishang:
Hi @gregorwolf , added the report zdemo_excel44 where this feature is used.
It's not a big thing but I encountered a requirement to have no empty lines ( specially CSV ) when the internal table to process was empty.

gregorwolf:
I've made a small correction on the test report and aded it to the main test report.

Issue #610, opened 2019-04-09
Error handling in ZCL_EXCEL_WORKSHEET->GET_TABLE


ivan-re:
The method ZCL_EXCEL_WORKSHEET-> GET_TABLE generates an error (Sy-subrc = 8) if there is no content in the last row of the last column of Excel.
Please extend the coding with this line (102):
                  **AND lv_rc <> 8.** "rc is 8 when the last row contains cells with zero / no values

Context:
                me-> get_cell (
                  EXPORTING
                    ip_column = lv_actual_col "Cell Column
                    ip_row = lv_actual_row "Cell Row
                  IMPORTING
                    ep_value = lv_value "Cell Value
                    ep_rc = lv_rc "Return Value of ABAP Statements
                ).
                IF lv_rc <> 0
                  AND lv_rc <> 4 "No found error, zero / no value in cell
                  **AND lv_rc <> 8**. "rc is 8 when the last row contains cells with zero / no values
                  lv_errormessage = | {'Error at reading field value (Col:' (007)} {lv_actual_col} {'Row:' (005)} {lv_actual_row} |.
                  zcx_excel => raise_text (lv_errormessage).
                ENDIF.

[GET_TABLE.txt](https://github.com/sapmentors/abap2xlsx/files/3058112/GET_TABLE.txt)

gregorwolf:
Thank you for letting us know. The perfect solution would be if you fork the project, make the correction and send a pull request.

PR #620, opened 2019-09-04
Fix Issue #610 (Error if last column of last row is empty)

striezl:
Ran into the troubles as described in issue  #610 an pasted in solution as suggested there.

The value of variable "lv_rc" which is checked here comes from a previous READ TABLE statement in method get_cell(). Apparently ABAP sets the return value to "8", when the last line of a internal
 table has been reached with TABLE KEY search, even though from my point of view the ABAP documentation (see attachment) is not 100% precise here. Nevertheless i can't see any reason why "8" (reached last row) should be treated different from "4" (did not last row) in this context. 

In fact from my point of view the whole exception handling in method get_table after me->get_cell()... (lines 96 et seq.) is (or has become) pointless and could be removed. But for now I wanted to keep the  change as minimal as possible.


larshp (approved):
Looks good to me, thanks

PR #627, opened 2019-09-20
range conversion fixed for range #REF!

d-schaaf:
working example: '2-ра акция UBP1'!$A$1:$T$2951
caused error: Sheet3!#REF!

Problem fixed by ignoring invalid ranges when autofilters are applied in zcl_excel_reader_2007->load_workbook

check for comments
" begin Dennis Schaaf
" end Dennis Schaaf


gregorwolf:
Hi Dennis,

can you please check that you have the latest abapGit version in your system and please import the most current version of abap2xlsx. Currently a lot of changes are in your commit that are just changes in the case. So perhaps avoid using pretty printer here.

CU
Gregor

d-schaaf:
Hi Gregor,

yeah, sorry for pretty printer ☹
That’s why I added some comments to find my changes easier.
ABAP2XLSX has been forked just before I made the changes, so it should be the most current version.
ABAPGIT I’m not sure if it’s up2date or not, I can try to refresh. Just a question: Why is ABAPGIT important to be current version?

However, do I have to commit my changes again, open an issue or how to proceed with such problems?

Regards,
Dennis

PR #657, opened 2020-04-17
ATC check cleared

d-schaaf:
We found ATC check "Empty SELECT... ENDSELECT loop" when selecting the date format from USR01.

- removed SELECT * from USR01
- removed empty SELECT...ENDSELECT on URS01 and changed into SELECT SINGLE


d-schaaf:
Fixed another ATC issue in ZCL_EXCEL_READER_2007: ZIF_EXCEL_READER~CAN_READ_FILE was implemented but method no more defined in the interface. Method implementation has been deleted

gregorwolf:
LGTM

PR #663, opened 2020-04-28
Reading corrupted xlsx file

JuvickAul:
A cell contained text in the original template. It was lated deleted. But apparently a link to SharedStrings from the cell was not removed. An exception occurred after downloading and opening the file via ZCL_EXCELL_READER_2007 class. All operations until ZCL_EXCEL_READER_2007 were done using other tools.


gregorwolf (approved):
LGTM

PR #665, opened 2020-05-14
Update zcl_excel_converter.clas.abap

K-Wagner:
Converting an ALV table that includes a color column failed because deep tables are not supported. The check that was addes prevents an error.

### DIFF
--- a/src/zcl_excel_converter.clas.abap
+++ b/src/zcl_excel_converter.clas.abap
@@ -1415,7 +1415,6 @@ method LOOP_SUBTOTAL.
   LOOP AT wt_fieldcatalog ASSIGNING <fs_sfcat>.
     l_row_int = i_row_int.
     l_col_int = i_col_int + <fs_sfcat>-position - 1.
-    
 * Freeze panes
     IF <fs_sfcat>-fix_column = abap_true.
       ADD 1 TO r_freeze_col.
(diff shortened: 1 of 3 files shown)


gregorwolf:
Hi @K-Wagner,
is there any sample report that could be adjusted to replicate this issue first and then use your correction?

K-Wagner:
I've tried to reproduce the error using ZDEMO_EXCEL32. When run interactively the color column is excluded in ZCL_EXCEL_COMMON~DESCRIBE_STRUCTURE, no problem there. I recalled I wanted to reproduce Excel files in batch mode also using an ALV Layout. There the problem begins. I enhanced ZDEMO_EXCEL32 to produce the output also when called from ZDEMO_EXCEL. Right now the color information is missing. Also there is made no column selection based on an ALV layout. Working on that.

AndreaBorgia-Abo:
#781 was merged and resulted in some conflicts, sorry for the inconvenience.

sandraros:
Hi @K-Wagner. I tried to reproduce but couldn't understand what problem you exactly faced. If I run the demo program you have proposed, I don't see any failure. The catalog returned by SALV in my 7.52 version doesn't contain the `COLOR_CELL` table type component you have added to contain the colors of cells. To continue the discussion, could you create an issue with details about how to reproduce? Thank you.

sandraros:
I propose to close this PR for now. If there is still an issue, I propose that first an issue is opened which describes the exact reproduction steps. Thank you.

For history, here's what I obtain when I run the adapted version of ZDEMO_EXCEL32 provided in this PR after I click the button "Excel converter":
image

PR #712, opened 2020-11-19
Fix in demo 11: Field-symbol usage

StefanRutzmoser:
Do not rely on IF ... IS ASSIGNED for a field symbol inside a LOOP


gregorwolf (approved):
LGTM

PR #734, opened 2021-02-09
Repeated column formulas

sandraros:
Fix to implement the feature described in #723


sandraros:
Replaced with newest #843

Issue #740, opened 2021-02-13
Impossible to define a default style for the whole row


sandraros:
It's currently possible to define a style for a column, but not for a row.

Moreover, if an Excel file is read, the default style of the row is currently ignored and cannot be obtained. Note that if a cell value is defined in this row, Excel indicates the row default style in the cell, so the formatting of this cell correctly reflects the style, but not the other cells in the row, but in that case there is still an issue for the other cells, like the background color for instance.

For columns, it's possible and an example is given in program `ZDEMO_EXCEL36`:
```
  lo_column = lo_worksheet->get_column( 'B' ).
  lo_column->set_column_style_by_guid( ip_style_guid = lv_style_times11_guid ).
```
The class ZCL_EXCEL_COLUMN contains the methods `set_column_style_by_guid` and  `get_column_style_guid`.

Is it possible to add the methods `set_style_by_guid` and  `get_style_guid` in class `ZCL_EXCEL_ROW`?

NB: currently, the workaround is to define all the cells of the row with a given style.
NB: the issue was detected in #618 and #722

AndreaBorgia-Abo:
Yes, I agree it would make sense to have them.

The only possible issue I see is when both setter methods are called on the same sheet with conflicting formats: however, just like with Excel, the last one wins.

Personally, I would extend ZDEMO_EXCEL36 to showcase this feature instead of creating a new one but that's really up to you.

Issue #722, opened 2021-01-25
Wrongly changed worksheet dimension


chemik3:
When I open file input.xlsx with zcl_excel_reader_2007 and immediately write with zcl_excel_writer_2007. New file had changed format, from row 17 is changed font size from 11 to 12. In source file sheet1.xml are tags which generated file don't have:

https://imgur.com/3KyEdyp

For test you can use test program: https://drive.google.com/file/d/18Z8mCteUvR9v15JB49ALtkKXczmgPLg0/view?usp=sharing 
import xlsx file: https://drive.google.com/file/d/1EhjF5NhfUlRMu3Sao8_A9jnBaqE7v3r-/view?usp=sharing

sandraros:
Problem reproduced. For information, I can see that the UsedRange of the sheet has changed. Original UsedRange (press Ctrl+End) is AY784 (`<dimension ref="A1:AY784"/>`), while the generated file UsedRange is AY16 (`<dimension ref="A2:AY16"/>`). In the original one, I see that AY784 has font size 11, but the next rows have font size 12.

sandraros:
Analysis...
So, the original XLSX file is like this one, with a lot of formatting, different row heights, different styles with different font sizes, but the cells are empty:
![image](https://user-images.githubusercontent.com/34005250/107852695-8254e300-6e12-11eb-89bc-90f4ee02bedb.png)

In the XML file of the original sheet, we can see that the last row with cells inside is the row 16, but there are extra rows up to row 784 with style, custom format and height:
![image](https://user-images.githubusercontent.com/34005250/107852659-59345280-6e12-11eb-9ab4-934000855c0e.png)

When the custom program calls the method READ_FILE, after that the SHEET_CONTENT attribute of the worksheet object has last cell at position AY16:
![image](https://user-images.githubusercontent.com/34005250/107852742-bcbe8000-6e12-11eb-92d4-a1eb3b286313.png)

and the ROWS attribute of the worksheet object has only 3 rows (1):
![image](https://user-images.githubusercontent.com/34005250/107852774-e4ade380-6e12-11eb-8c91-00b8d19f5a9c.png)
(1) because the XML attribute "customHeight" is "1" (ECMA-376 says "1 if the row height has been manually set."):
![image](https://user-images.githubusercontent.com/34005250/107853111-67836e00-6e14-11eb-97fa-080975655e3b.png)
[...]

PR #741, opened 2021-02-13
Wrongly changed file size #722

sandraros:
When reading a worksheet, keep empty rows which have specific height or style #722 (should be completed with correction of #740)


AndreaBorgia-Abo:
Reviewed the report, checked the contents of the sheet1.xml file before and after importing the changes in my demosystem, also checked the xlsx files visually. Looks good to me.

PR #752, opened 2021-03-22
use ZABAP2XLSX message class

AndreaBorgia-Abo:
Not quite sure about the meaning of "message invalid" because I could find the corresponding classes on my 7.52SP04 system but this should take care of #659.


AndreaBorgia-Abo (commented):
Changes are as expected

AndreaBorgia-Abo:
@larshp could you please merge this?

AndreaBorgia-Abo:
thanks

PR #774, opened 2021-07-10
remove usage of OS_BOOLEAN ( WIP #692 )

AndreaBorgia-Abo:
I could find only one use of the OS_BOOLEAN DTEL in all of abap2xlsx, precisely in zdemo_excel_outputopt_incl where it is used for the return value of a standard class method, cl_send_request->send.

"send" calls "release" which calls "submit" which finally sets e_sent_to_all to X unless one recipient has issues, in which case it is cleared: for this reason, replacing "IS INITIAL" with "= abap_false" is incorrect and the test should be inverted.

WIP #692


larshp (approved):
LGTM 👍

PR #786, opened 2021-07-31
Fix for Issue #783 - Conditional Formatting for Cells that "Begin With" a certain string - Third Attempt

hardyp:
This time I changed the code using Eclipse. When I look at the DIFF inside abapGit then I only see what I expect to see i.e. a few lines of added code.


AndreaBorgia-Abo:
Adding myself as reviewer so I won't forget to try it once I'm back from holidays in a couple of weeks.
If anyone has a chance to review it before then, be my guests :)

sandraros:
@AndreaBorgia-Abo I have some free time this week. Is it fine if I continue working on this issue/fix?

hardyp:
Fine with me. I won't have to time to come back to this for a month at least....

AndreaBorgia-Abo:
Also fine with me, I'm a bit backlogged because of real life :)
I'll try to find the time for tests, if needed, @sandraros

sandraros:
I would like your opinion about doing the things a little bit differently. Sorry for the big message.

Situation before a fix:
1. all the 4 text-specific operators don't work (beginswith, contains, etc.)
2. we have 1 confusing rule constant "c_rule_containsText" which in fact does nothing.
3. there are 12 constants prefixed "c_operator", 7 of them are for "c_rule_cellis", 1 is not used ("none"), and 4 others are the text-specific operators; if the developers want to use the latter 4 they would surely think they should use "c_rule_cellis" because of their prefix.

The proposed solution (edit: the proposed solution by Paul) is to define both a new rule "beginswith" and operator "beginswith", and continue using "mode_cellis". It's coherent with Office Open XML format, but I see possible pain points for the developer:
- up to now "c_rule_xxxx" is always paired with a "mode_xxxx" attribute and here we break the rule (there's no "mode_beginswith"), so it can be confusing
- the developer must enter the same operator in two fields (although it's what Office Open XML does). We could restrict to one field by deducing the other field (3 operators have identical values, the 4th one has different values "notContains"/"notContainsText").
- pain point 3 is still here (the developer will probably try to use "c_rule_cellis" instead of "beginswith")

I would like to extend the solution to fix all 4 text-specific operators at once, and make it easier for the developer to choose the right options:
- create the constant "c_rule_textfunction" and the attribute "mode_textfunction" with components "textfunction" (whose possible values are the 4 text-specific functions) and "text" (that will prevent confusing with respectively "operator" and "formula" in the attribute "mode_cellis")
- rename the prefix of the 4 text-specific operators, from "c_operator_" to "c_textfunction_" so that to reduce the risk of people using them in "mode_cellis-operator".
- Example (that could be added to `ZDEMO_EXCEL27`:
  ```
  lo_style_cond = lo_worksheet->add_new_style_cond( ).
  lo_style_cond->rule          = zcl_excel_style_cond=>c_rule_textfunction.
  ls_textfunction-text         = 'C'.
  ls_textfunction-textfunction = zcl_excel_style_cond=>c_textfunction_beginswith.
  ls_textfunction-cell_style   = lv_style_2_guid.
  lo_style_cond->mode_textfunction = ls_textfunction.
  lo_style_cond->priority     = 2.
  lo_style_cond->set_range( ip_start_column  = 'C'
                            ip_start_row     = 2
                            ip_stop_column   = 'C'
                            ip_stop_row      = 2 ).
  ```
You can see the proposed code here: https://github.com/sandraros/abap2xlsx/tree/Fix-%23783-conditional-formatting

sandraros:
fyi demo program `ZDEMO_EXCEL27` updated in [my proposed solution](https://github.com/sandraros/abap2xlsx/tree/Fix-%23783-conditional-formatting) to demonstrate 11 formatting rules (equals, not equals, ..., contains, begins with, ...)

AndreaBorgia-Abo:
So, the demo works fine, at least with LibreOffice on my personal laptop.

Regarding the pain points:

1. I am not sure I understand your point on consistency between "rule" and "mode", though, since post changes we'd have "mode" for all "rule" constants which matter:
c_rule_cellis -> mode_cellis
c_rule_databar -> mode_databar
c_rule_expression -> mode_expression
c_rule_iconset -> mode_iconset 
c_rule_colorscale -> mode_colorscale
c_rule_none -> pointless, except for consistency
c_rule_textfunction -> mode_textfunction
c_rule_top10 -> mode_top10
c_rule_above_average -> mode_above_average

1. Developer must enter operator in two fields: where can I see this problem? ZDEMO_EXCEL27 looks fine. If you meant that you first need to ask for a text function (as rule) and then specify which of the 4 candidates, we can live with that, honestly.

1. I would use c_rule_textfunction so I think your proposal is good.

TLDR looks good to me, I guess @sandraros could push the changes to @hardyp 's branch and then we could merge this.

sandraros:
@AndreaBorgia-Abo Thank you for reviewing. In fact, your points 1 and 2 were in the original code proposed by Paul (so I fixed my comment above to say "The proposed solution **by Paul**"). As you agree with my proposal, let me push the changes to Paul's branch.

sandraros:
@hardyp I pushed my proposal to your branch. Please let us know what you think. Thanks! @AndreaBorgia-Abo fyi

AndreaBorgia-Abo (approved):

AndreaBorgia-Abo:
@hardyp are you ok with the end result? Any comments?
May we merge this?

hardyp:
I have been following this although I have been too busy to actively intervene. It all looks good to me so please go ahead.

PR #793, opened 2021-09-11
Bind table error if name like A1 (bug #733)

### LINKED ISSUE #733: BIND_TABLE with table_name corresponding to a cell address
This code leads to an error in Excel when opening the file:
```
DATA: gc_save_file_name TYPE string VALUE 'test.&'.
START-OF-SELECTION.
  SELECT * FROM scarr INTO TABLE @DATA(scarr).
  lo_worksheet->bind_table(
    EXPORTING
      ip_table            = scarr
      is_table_settings   = VALUE zexcel_s_table_settings(
                              table_name       = 'XFD1'  " <====== unfortunately this name is also a cell reference
                              top_left_column  = 'A'
                              top_left_row     = 1 ) ).
```
Excel error (Excel 365 MSO 16.0 64 bits):
image
Yes -> 
image

I found out the reason in this [Microsoft document - Important notes for names](https://support.microsoft.com/en-us/office/rename-an-excel-table-fbf49a4f-82a3-43eb-8ba2-44d21233b114), it's because XFD1 is a cell reference:
> Use valid characters — Always begin a name with a letter, an underscore character (_), or a backslash (\). Use letters, numbers, periods, and underscore characters for the rest of the name.
> Exceptions: You can’t use "C", "c", "R", or "r" for the name, because they’re already designated as a shortcut for selecting the column or row for the active cell when you enter them in the Name or Go To box.
> **Don’t use cell references — Names can’t be the same as a cell reference, such as Z$100 or R1C1.**
> Don’t use a space to separate words — Spaces cannot be used in the name. Consider how you can write the name using no spaces. Or, use an underscore character (_) or a period (.) as word separators. Examples: DeptSales, Sales_Tax or First.Quarter.
> Maximum 255 characters — A table name can have up to 255 characters.
> Use unique table names — Duplicate names aren’t allowed. Excel doesn’t distinguish between upper and lowercase characters in names, so if you enter “Sales” but already have another name called “SALES" in the same workbook, you’ll be prompted to choose a unique name.

The name "XFE1" will work because the column "XFD" is the last possible one (16384th column), so "XFE1" is not a cell reference.

The name "C" also leads to an error as explained above. I guess that the other mentioned rules are correct and should be checked in the correction.

sandraros:
Fix #733


AndreaBorgia-Abo:
Same as #839 : need Excel proper to test results.

src/zcl_excel_writer_2007.clas.abap
```diff
@@ -7283,7 +7283,18 @@ CLASS zcl_excel_writer_2007 IMPLEMENTATION.
 
     FIND ALL OCCURRENCES OF REGEX '[^_a-zA-Z0-9]' IN io_table->settings-table_name IGNORING CASE MATCH COUNT lv_match.
     IF io_table->settings-table_name IS NOT INITIAL AND lv_match EQ 0.
-      lv_table_name = io_table->settings-table_name.
+      " Name rules (https://support.microsoft.com/en-us/office/rename-an-excel-table-fbf49a4f-82a3-43eb-8ba2-44d21233b114)
+      "   - You can't use "C", "c", "R", or "r" for the name, because they're already designated as a shortcut for selecting the column or row for the active cell when you enter them in the Name or Go To box.
+      "   - Don't use cell references — Names can't be the same as a cell reference, such as Z$100 or R1C1
+      IF ( strlen( io_table->settings-table_name ) = 1 AND io_table->settings-table_name CO 'CcRr' )
```
AndreaBorgia-Abo, line 7289: Minor question: on line 7289 I have a warning. Scratch that, there are many such warnings :)
`The exception ZCX_EXCEL is not caught or declared in the RAISING clause of "CREATE_XL_TABLE". `
Anyway, the PR does the job: issue was present before and is gone afterwards.

AndreaBorgia-Abo (approved):

src/zcl_excel_writer_2007.clas.abap
sandraros: Yeah. Could be another PR to get rid of these syntax check messages (in many classes I think).

PR #821, opened 2021-10-03
Fix #547 autofilter messed up if several sheets

### LINKED ISSUE #547: Autofilter does not working correctly when using a template

sandraros:
Fix #547


AndreaBorgia-Abo (approved):
LGTM.
Either here or in the next PR, ZEXCEL_DEMO37 is missing some selection texts since the dawn of time, apparently: could you add them?

sandraros:
@AndreaBorgia-Abo and @larshp Thanks so much! I have created #826 for fixing ZDEMO_EXCEL37 another time.

PR #822, opened 2021-10-03
Fix #546 autofilter on row>1 and one-row height

### LINKED ISSUE #546: Autofilter on the wrong Row when you have an header line

sandraros:
Fix #546


AndreaBorgia-Abo (approved):
LGTM

Issue #807, opened 2021-09-24
Method BIND_ALV creates Excel file which ignores the ALV row filter


sandraros:
To reproduce:

1. Run demo program `ZDEMO_EXCEL32`
2. Filter manually the rows with any criteria
3. Click the button "Excel bind" to download the ALV to an Excel file
4. The Excel file shows all the rows, the filtered rows are shown

NB: this issue comes from #745

PR #837, opened 2021-10-07
Fix #745 Bind ALV with non-displayed SALV data

### LINKED ISSUE #745: Report to save Excel from SALV (without displaying it): what is the recommended way?
Hello, 

This use case is quite common: I want to enhance an SALV report with the ability to download an Excel *instead of* displaying it (similar to the output radiobutton option in the demo programs).

I don't believe I'm the first one to try it and I would like to know what everyone do.

### Best attempt
My best attempt is the following. 

```ABAP
sy-batch = abap_true.
data(excel) = new zcl_excel( ).
data(worksheet) = result->get_active_worksheet( ).
worksheet->bind_alv(
    io_alv      = salv
    it_table    = table  ).

sy-batch = abap_false. 
data(writer)     = cast zif_excel_writer( new zcl_excel_writer_2007( ) ).
data(excel_xstring) = writer->write_file( excel ).    
```
advantages:
  * I avoid the infamous ```OBJECTS_OBJREF_NOT_ASSIGNED``` dump on ```CL_SALV_TABLE->GET_METADATA```
  * the excel columns match what is defined in layout
  * subtotals are displayed

inconvenient:
  * I don't have filter

### second best
```ABAP
cl_salv_bs_runtime_info=>set(
  exporting
    display        = abap_false
    metadata       = abap_true
    data           = abap_true ).
salv->display( ).

data(excel) = new zcl_excel( ).
data(worksheet) = result->get_active_worksheet( ).
worksheet->bind_alv(
    io_alv      = salv
    it_table    = table  ).

data(writer)     = cast zif_excel_writer( new zcl_excel_writer_2007( ) ).
data(excel_xstring) = writer->write_file( excel ).
```
Inconvenient:
  * I don't have neither filter, nor hidden columns or subtotals from the layout (From what I understand, the layout options have not been applied to the internal SALV fieldcatalog when ```ZCL_EXCEL_CONVERTER_SALV_TABLECP=>LOAD_DATA``` does not take layout).

Thanks,
PS: I promise I'll PR a doc page with the outcome of the discussion.

sandraros:
Fix #745


AndreaBorgia-Abo:
Note to self (mostly): to test this PR use Excel only, Librecalc won't do; the columns are hidden as expected and the sum is set but the function code is not recognized and an error is displayed.

src/zcl_excel_converter_salv_table.clas.abap
```diff
@@ -253,4 +269,32 @@ CLASS zcl_excel_converter_salv_table IMPLEMENTATION.
...
+    DATA: lr_s_type_runtime_info TYPE REF TO data.
+    FIELD-SYMBOLS: <ls_type_runtime_info> TYPE any,
+                   <lv_display>           TYPE any,
+                   <lv_data>              TYPE any.
+
+    rv_result = abap_false.
+    TRY.
+        CREATE DATA lr_s_type_runtime_info TYPE ('CL_SALV_BS_RUNTIME_INFO=>S_TYPE_RUNTIME_INFO').
+        ASSIGN lr_s_type_runtime_info->* TO <ls_type_runtime_info>.
+        CALL METHOD ('CL_SALV_BS_RUNTIME_INFO')=>('GET')
```
AndreaBorgia-Abo, line 284: Only one question: why use the dynamic call syntax when the classname / method are both fully determined?

AndreaBorgia-Abo (approved):
Other than that one question, the code works and looks good.
I'll wait for an answer but to me the PR can otherwise be merged.

src/zcl_excel_converter_salv_table.clas.abap
sandraros: It's because I don't know if this class and this method exist in all ABAP versions, and currently abap2xlsx should be made to work starting with ABAP 7.02. The solution compiles and works with 7.52 but I don't know before.

mahlzeit1948:
Hello, has this been solved? I am running the latest version of abap2xlsx and I am still encountering the issue that layout filters are not taken into account...no matter whether the ALV is being displayed or not.

AndreaBorgia-Abo:
Could you please open another issue? It'd be great if you could provide a sample program showcasing the issue.
We will then have a look and reference the original issue if that turns out to be the case.

AlexandreHT:
#807 (Method BIND_ALV creates Excel file which ignores the ALV row filter) is still open

AndreaBorgia-Abo:
@AlexandreHT thanks for catching that, I had completely forgotten about it :(

PR #839, opened 2021-10-07
Fix #836 new feature "Ignored Errors"

### LINKED ISSUE #836: Feature request for Ignored Errors
In Excel, there is a feature to ignore the "errors" (can also be warnings). Currently, this feature doesn't exist in abap2xlsx. Is it possible to add it? It's quite a minor feature, essentially aesthetic.

# Example 1
```
CONSTANTS gc_save_file_name TYPE string VALUE 'issue-836.xlsx'.
START-OF-SELECTION.
  SELECT carrid, connid, fldate, bookid FROM sbook UP TO 3 ROWS
  INTO TABLE @DATA(gt_sbook).
  TRY.
      DATA(excel) = NEW zcl_excel( ).
      DATA(worksheet) = excel->get_active_worksheet( ).
      worksheet->bind_table( ip_table = gt_sbook ).
      lcl_output=>output( excel ).
    CATCH cx_root INTO DATA(lx).
      MESSAGE lx TYPE 'I' DISPLAY LIKE 'E'.
  ENDTRY.
```
- Result of the program above:
  image
- Manual fix:
  image
- Expected result (what we should be able to do with abap2xlsx):
  image
 
# Example 2
See [file attached](https://github.com/sapmentors/abap2xlsx/files/7305277/abap2xlsx_ignored_errors.xlsx) (NB: the error indicator may happen in many other situations):
- Create this worksheet containing a Table, with Column2 containing formula `=[@Column1]`, and force a text in cell `B4`, it then shows an error indicator, see here:
  image
- Click the "`!` losange" icon, and choose "Ignore Error":
  image
- It's not shown any more as an error:
  image

NB: this issue is created to simplify the solution provided in #734 which is about adding the Excel feature "Calculated Columns".

sandraros:
Fix #836


AndreaBorgia-Abo:
Using code from Example1 or the file attached in Example2 with Libreoffice I cannot get any errors even with the current master. I will have to try again using the company laptop (with real Excel), middle of next week. (edit: make that next friday)

AndreaBorgia-Abo:
My mistake: manual merging with the GH interface did not go down well. Will fix tomorrow morning, along with review. Sorry.

src/zcl_excel_reader_2007.clas.abap
```diff
@@ -232,6 +232,12 @@ CLASS zcl_excel_reader_2007 DEFINITION
         !it_external_hyperlinks TYPE gtt_external_hyperlinks
       RAISING
         zcx_excel .
+    METHODS load_worksheet_ignored_errors
+      IMPORTING
+        !io_ixml_worksheet TYPE REF TO if_ixml_document
+        !io_worksheet      TYPE REF TO zcl_excel_worksheet
```
AndreaBorgia-Abo, line 238: I'm really torn between objecting to its (over)use in new code and letting it go because it is after all inconsequential.
I assume you went along [with the flow](https://github.com/sapmentors/abap2xlsx/issues/773#issuecomment-877645825), right? Like, you didn't really use SE24, did you? :)

AndreaBorgia-Abo (approved):

AndreaBorgia-Abo:
LGTM

AndreaBorgia-Abo:
I wanted to trim the commit messages a bit and ended up creating a revert commit :/
Fuggedaboutit! The code is good, the log should have been cleaner but it's ok.

src/zcl_excel_reader_2007.clas.abap
sandraros: Yes, I use Eclipse. The exclamation mark is not something precisely discussed in the coding guidelines, my idea is to make the source code coherent with the rest of code as far as possible, so I have typed them. We may remove the exclamation marks if you prefer, no problem for me. Maybe we can do an additional PR each time we do a fix in a class, to remove all the exclamation marks (we could track the list of processed classes by creating a new issue).

AndreaBorgia-Abo: You're right, let's move the discussion back there: this way, we can gather feedback and include it in the guidelines.

PR #842, opened 2021-10-08
Remove dtel char1

AndreaBorgia-Abo:
Couple of things need a closer look:
1. lv_last_allowed_char receives a 2-byte value, it should have been CHAR2 even before the DDIC cleanup
1. x_references / y_references unused : potentially a breaking change but I just don't see people accessing random attributes that the class itself doesn't care about


AndreaBorgia-Abo (commented):
changes now look as intended, sorry for initial mixup

src/zcl_excel_converter.clas.abap
```diff
@@ -104,11 +104,11 @@ CLASS zcl_excel_converter DEFINITION
...
-    CONSTANTS c_type_hdr TYPE char1 VALUE 'H'.              "#EC NOTEXT
-    CONSTANTS c_type_str TYPE char1 VALUE 'P'.              "#EC NOTEXT
-    CONSTANTS c_type_nor TYPE char1 VALUE 'N'.              "#EC NOTEXT
-    CONSTANTS c_type_sub TYPE char1 VALUE 'S'.              "#EC NOTEXT
-    CONSTANTS c_type_tot TYPE char1 VALUE 'T'.              "#EC NOTEXT
+    CONSTANTS c_type_hdr TYPE c VALUE 'H'.              "#EC NOTEXT
+    CONSTANTS c_type_str TYPE c VALUE 'P'.              "#EC NOTEXT
+    CONSTANTS c_type_nor TYPE c VALUE 'N'.              "#EC NOTEXT
+    CONSTANTS c_type_sub TYPE c VALUE 'S'.              "#EC NOTEXT
+    CONSTANTS c_type_tot TYPE c VALUE 'T'.              "#EC NOTEXT
```
sandraros, line 111: I think it's recommended in ABAP doc to write `TYPE c LENGTH 1` explicitly. Don't know if Clean Code confirms.

src/zcl_excel_converter.clas.abap
```diff
@@ -215,7 +215,7 @@ CLASS zcl_excel_converter DEFINITION
         VALUE(r_function_number) TYPE int1 .
     METHODS get_style
       IMPORTING
-        !i_type        TYPE char1
+        !i_type        TYPE c
```
sandraros: Here `TYPE c` means the generic form, it could be any C data object of any length, so we'd regress with original code. We should declare a type explicitly `TYPES ty_style_type TYPE c LENGTH 1.`

src/zcl_excel_converter.clas.locals_def.abap
```diff
@@ -27,7 +27,7 @@ TYPES:   BEGIN OF ts_subtotal_rows,
          tt_subtotal_rows TYPE HASHED TABLE OF ts_subtotal_rows WITH UNIQUE KEY row_int.
 
 TYPES:   BEGIN OF ts_styles,
-            type         TYPE char1,
+            type         TYPE c LENGTH 1,
```
sandraros: `TY_STYLE_TYPE` could be used, based on my above comment.

src/zcl_excel_drawing.clas.abap
```diff
@@ -133,14 +131,14 @@ CLASS zcl_excel_drawing DEFINITION
     DATA type TYPE zexcel_drawing_type VALUE type_image. "#EC NOTEXT .  .  .  .  .  .  .  .  .  .  . " .
     DATA index TYPE string .
     DATA anchor TYPE zexcel_drawing_anchor VALUE anchor_one_cell. "#EC NOTEXT .  .  .  .  .  .  .  .  .  .  . " .
-    CONSTANTS c_media_source_www TYPE char1 VALUE 1.        "#EC NOTEXT
-    CONSTANTS c_media_source_xstring TYPE char1 VALUE 0.    "#EC NOTEXT
-    CONSTANTS c_media_source_mime TYPE char1 VALUE 2.       "#EC NOTEXT
+    CONSTANTS c_media_source_www TYPE c VALUE 1.        "#EC NOTEXT
+    CONSTANTS c_media_source_xstring TYPE c VALUE 0.    "#EC NOTEXT
+    CONSTANTS c_media_source_mime TYPE c VALUE 2.       "#EC NOTEXT
```
sandraros, line 136: `TYPE c LENGTH 1`

src/zcl_excel_drawing.clas.abap
```diff
@@ -133,14 +131,14 @@ CLASS zcl_excel_drawing DEFINITION
...
     DATA guid TYPE guid_16 .
     DATA media TYPE xstring .
     DATA media_key_www TYPE wwwdatatab .
     DATA media_name TYPE string .
-    DATA media_source TYPE char1 .
+    DATA media_source TYPE c .
```
sandraros, line 141: `TYPE c LENGTH 1`

src/zcl_excel_writer_huge_file.clas.abap
```diff
@@ -51,7 +51,7 @@ CLASS zcl_excel_writer_huge_file IMPLEMENTATION.
       END OF ts_root.
 
     DATA:
-      lv_last_allowed_char TYPE char1,
+      lv_last_allowed_char TYPE c LENGTH 2,
```
sandraros: Not sure. I think it should be `TYPE c LENGTH 1`. In Unicode, hexa `FFFD` means 1 characters `U+FFFD`.

AndreaBorgia-Abo: I did ask myself about the unicode char vs byte issue but a quick test shows that the return of cl_abap_conv_in_ce=>uccp is SYCHAR02, basic type C(2)

sandraros: Yes it can be possibly 2 characters, for instance for emoji (and many others), but it would work with 1 character for FFFD. If we have a variable of 2 characters, it's confusing when we look at the RegEx because it's part of an interval (from U+0020 to U+FFFD), it's weird letting the developer think that there could be 2 characters in the high part of the interval. Maybe adding a comment to explain that it's 1 character on purpose?

src/zcl_excel_converter.clas.abap
AndreaBorgia-Abo: Clean Code has nothing about this specific issue and class zcl_excel_graph for example has the same generic "type c".
Constants being what they are, the size is determined at creation from the supplied value.

src/zcl_excel_converter.clas.abap
AndreaBorgia-Abo: You're right! Will fix.

src/zcl_excel_converter.clas.abap
sandraros: okay, I don't mind anyway ;-), `TYPE c` is fine. Just in case, it was about this kind of [mention](https://help.sap.com/doc/abapdocu_latest_index_htm/latest/en-US/index.htm?file=abapdata_simple.htm): "Hints For reasons of legibility, it is best to include all information and always use the addition LENGTH instead of parentheses to specify the length len."

src/zcl_excel_writer_huge_file.clas.abap
sandraros: I looked at the result returned by `UCCPI`, it's U+FFFD followed by U+0020 (space) so the code will behave the same whatever it's 1 or 2 characters. So technically it's fine.

src/zcl_excel_drawing.clas.abap
AndreaBorgia-Abo: Oddly enough, while in method parameters "c" is indeed any unspecified length, when it comes to variables and constants ABAP defaults to 1 when no length is given.

sandraros: It's default 1 for all 3 generic types C, N and X, and 8 for P. ("standard length" in tables of [ABAP types](https://help.sap.com/doc/abapdocu_latest_index_htm/latest/en-US/index.htm?file=abenbuilt_in_types_complete.htm)). And the same if you declare optional parameters with these 4 generic types and you don't pass an argument.

sandraros (approved):
LGTM 👍

PR #843, opened 2021-10-08
Fix #723 Calculated Columns - Replace PR 734

### LINKED ISSUE #723: Feature request for Excel tables: Calculated Column Formula

sandraros:
Fix #723. Currently I can do only the WRITER part, because the READER currently doesn't read the Tables at all, just its contents, i.e. the Table generated by BIND_TABLE doesn't have a counterpart in the READER.


AndreaBorgia-Abo:
Same as #839 : need Excel proper to test results.

AndreaBorgia-Abo:
The file generated by ZDEMO_EXCEL47 contains an error and was repaired by Excel 2013 (Windows) upon opening, this is how it looked like afterwards:
immagine

Here is the correction log:
```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<recoveryLog xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
	<logFileName>error096120_01.xml</logFileName>
	<summary>Rilevati errori nel file 'C:\temp\47_ColumnFormulas.xlsx'</summary>
	<removedRecords summary="Di seguito è riportato un elenco dei record rimossi:">
		<removedRecord>Record rimossi: Formula dalla parte /xl/worksheets/sheet1.xml</removedRecord>
		<removedRecord>Record rimossi: Tabella dalla parte /xl/tables/table1.xml (Tabella)</removedRecord>
	</removedRecords>
</recoveryLog>
```

sandraros:
Thanks. The problem is due to the Excel version, the feature works only from Excel 365 or Excel 2021 (https://support.microsoft.com/en-us/office/use-calculated-columns-in-an-excel-table-873fbac6-7110-4300-8f6f-aafa2ea11ce8).
image
image
image

Is it okay to add the required version in the title or should we do more? (maybe like defining a target Excel version while instantiating `ZCL_EXCEL` so that to raise an exception with the right message if the program attempts using some unsupported feature)
image

src/demos/zdemo_excel47.prog.xml
```diff
@@ -0,0 +1,23 @@
...
+    <DBNA>D$</DBNA>
+    <SUBC>1</SUBC>
+    <FIXPT>X</FIXPT>
+    <LDBNAME>D$S</LDBNAME>
+    <UCCHECK>X</UCCHECK>
+   </PROGDIR>
+   <TPOOL>
+    <item>
+     <ID>R</ID>
+     <ENTRY>abap2xlsx Demo: Table Calculated Columns</ENTRY>
```
AndreaBorgia-Abo: 2016 and above

sandraros:
Hmm sorry. Yes. And also valid for Excel 2013:
> Excel for Microsoft 365 Excel for Microsoft 365 for Mac Excel 2021 Excel 2021 for Mac Excel 2019 Excel 2019 for Mac Excel 2016 Excel 2016 for Mac Excel 2013 Excel 2010 Excel 2007 Excel for Mac 2011

The problem about column "C4." is due to `IFS` (https://support.microsoft.com/en-us/office/ifs-function-36329a26-37b2-467c-972b-4a39bd951d45):
> Excel for Microsoft 365 Excel for Microsoft 365 for Mac Excel for the web Excel 2021 Excel 2021 for Mac Excel 2019 Excel 2019 for Mac

The problem with columns "C6." and "C7." is due to `FILTER` (https://support.microsoft.com/en-us/office/filter-function-f4f7cb66-82eb-4767-8f7c-4877ad80c759):
> Excel for Microsoft 365 Excel for Microsoft 365 for Mac Excel for the web Excel 2021 Excel 2021 for Mac Excel for iPad Excel for iPhone Excel for Android tablets Excel for Android phones

What do you think?

AndreaBorgia-Abo:
Right, I asked a colleague and indeed Excel 2016 has no issues. The way I read the "calculated columns" list of versions is that it's pretty Mac-specific and I am using Windows.
I believe we could do both: edit the title (and maybe a note in the code with links) in this demo (to close this PR) and then open a bug to track the Excel version issue: I expect this kind of glitches could happen a lot but for a proper fix we'd need keep tables of applicability lists for each feature and have a master version switch against which to compare.
Unrelated question: why some PR have a branch on the main repository and some on yours? Just curious :)
Let me finish the review, we'll see...

sandraros:
Agreed for "applicability list". Let's close the PR first. Someone told me that Excel 2016 doesn't exist, it's the internal number, probably the official name is Excel 365 or Excel 2019 (https://en.wikipedia.org/wiki/Microsoft_Excel#Excel_2019,_Office_365_and_subsequent_(v16.0)). Each supported version is both for Mac and for Windows. I propose the text "(from Excel 365 or 2019)".
I now create a branch directly in abap2xlsx, the advantage is that it's automatically deleted when you merge the PR (it's what Lars does). I still have old PRs/branches in my own repository, I hope one day I can clean it.

AndreaBorgia-Abo (approved):

AndreaBorgia-Abo:
Hmm, I prefer to keep branches on my own repo to minimize the risk of pushing something to master outisde the PR channel.
Crucially, I also get a prompt after merging to delete the branch here on GH. At the next pull it will be gone from the demo system as well.
Anyway, merging :)

PR #845, opened 2021-10-09
remove uses of CHAR2 ( WIP #692 )

AndreaBorgia-Abo:
both DOMA and DTEL


AndreaBorgia-Abo (commented):
changes are as expected

sandraros (approved):
LGTM 👍

PR #846, opened 2021-10-09
remove uses of CHAR20 ( WIP #692 )

AndreaBorgia-Abo:
both DTEL and DOMA


AndreaBorgia-Abo (commented):
changes look as expected

sandraros (approved):
LGTM 👍

PR #855, opened 2021-10-16
remove usage of CHAR3 ( WIP #692 )

AndreaBorgia-Abo:
got a warning on activation:
immagine
(background: I tend to use SE11 for DDIC and Eclipse for the rest, I find that limiting the search area by $Z* as package works well in GUI but not so much via Eclipse)


sandraros (approved):
LGTM 👍

AndreaBorgia-Abo:
Back to the topic of tree location, now I see: since the tree is in my fork, I get a prompt but no automatic deletion. All told, I can live with that, all good :)

PR #856, opened 2021-10-16
remove usage of CHAR5 ( WIP #692 )

AndreaBorgia-Abo:


AndreaBorgia-Abo (commented):
changes are as expected

sandraros (changes requested):
Is it worth "creating a new include" in `zcl_excel_writer_2007" for just defining the type `tv_charbool`. Why not defining it as a private type? I personally use the local definitions only for local classes/interfaces used in signatures of private methods.

NB: I just discovered that the [github search is very limited](https://docs.github.com/en/search-github/searching-on-github/searching-code). [github doesn't find "char5"](https://github.com/sapmentors/abap2xlsx/search?q=CHAR5&type=code) in all files 😒 (see below screenshot). I will have to download the files and scan by myself as an alternative to find out all usages easily...
image

AndreaBorgia-Abo:
@sandraros perhaps not, I saw the aptly named "class local types" tab in Eclipse and went along with it. I like your suggestion, I'll change this.
To be extra sure, when it comes to where-used, I first cross-check GUI (which supports search areas as $Z*) and Eclipse (which apparently only takes a full package name) and then also run a recursive grep (I keep a local clone because upstream merging using git cmdline is a piece of cake).

sandraros:
Nice. I'll try your SAP GUI/Eclipse tips for the where-used list, I didn't use them until now.

sandraros (approved):
LGTM 👍

PR #865, opened 2021-11-01
solves part of #859 - split tests for is_cell_in_range

AndreaBorgia-Abo:


AndreaBorgia-Abo (commented):
changes are as expected

PR #886, opened 2021-11-07
solves part of #859 - split tests for shift_formula

AndreaBorgia-Abo:


AndreaBorgia-Abo (commented):
changes are as expected, this should be the last part ( and, no, it isn't XD )

larshp (approved):

PR #910, opened 2021-12-10
Reader fails for workbook with chart

### LINKED ISSUE #790: ABAP2XLSX - Reading Spreadsheet with existing charts
I am using ABAP2XLSX to read an existing spreadsheet. The existing charts that are defined in this spreadsheet appear to get lost when I do this. 

I tested this with ZDEMO_EXCEL37. The source spreadsheet I read had existing charts. It is attached.  In the new spreadsheet that is created after I run the transaction, the charts are no longer there.

[VGV_UTILIZATION_81321.xlsx](https://github.com/sapmentors/abap2xlsx/files/7008490/VGV_UTILIZATION_81321.xlsx)

sandraros:
fix #790

### DIFF
--- a/src/zcl_excel_drawing.clas.abap
+++ b/src/zcl_excel_drawing.clas.abap
@@ -826,8 +826,8 @@ CLASS zcl_excel_drawing IMPLEMENTATION.
         node ?= node->find_from_name( name = 'lineChart' namespace = 'c' ).
         node2 ?= node->find_from_name( name = 'marker' namespace = 'c' depth = '1' ).
         IF node2 IS BOUND.
-        zcl_excel_reader_2007=>fill_struct_from_attributes( EXPORTING ip_element = node2 CHANGING cp_structure = ls_prop ).
-        lo_linechart->ns_markerval = ls_prop-val.
+          zcl_excel_reader_2007=>fill_struct_from_attributes( EXPORTING ip_element = node2 CHANGING cp_structure = ls_prop ).
+          lo_linechart->ns_markerval = ls_prop-val.
         ENDIF.
         node2 ?= node->find_from_name( name = 'smooth' namespace = 'c' depth = '1' ).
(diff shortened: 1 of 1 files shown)


larshp:
pretty print to fix the indentation?

sandraros:
Shouldn't we pretty print only after the review, so that the reviewer can check what has been exactly changed first, and after approval we add a commit with the pretty print? Now, the diff doesn't mess with the indentations and it shows the exact lines added. Even if there's not much changes, I prefer to trust the diff than my eyes. By the way, after that, we could do the pretty print on 8 source codes that I missed last time.

larshp:
well, then the reviewer would have to review the same code two times

plus ideally, the [indentation rule](https://rules.abaplint.org/indentation/) would be enabled, forcing the developer to pretty print before the review starts

sandraros:
I don't completely agree on this, the reviewer can easily do the pretty print. For this PR, as it's quite simple, let me add the commit with pretty print. But we had recently cases where a big section of code was (de-)indented because of a condition/loop changed around, and after the pretty print the diff was a mess. It's much easier to review when you are sure that only 2 lines are changed. I would agree to install the indentation rule only after we have refactored all the big methods of abap2xlsx (after that, if you have an abaplint rule to prevent big methods, that would be great). For information, I've got [this program](https://gist.github.com/sandraros/9282da1c7c39c944a87d843e49508651) to show instantly what is wrong with the pretty print in all abap2xlsx.

larshp:
a reviewer reviews changes, if there are additional changes these should also be reviewed

the reviewer should review the final state of a change

the reviewer might choose to review the change not looking at whitespace,
image

the codebase should be consistent, with rules enabled to enforce the preferred style

larshp:
avoid long methods: https://rules.abaplint.org/method_length/

sandraros:
Thanks 👍, understood and fully agree now! 😉

larshp:
and having stuff enabled, we can focus time on the actual changes 😅

PR #912, opened 2021-12-10
# Huge writer error height/hight and fit to page

### LINKED ISSUE #553: Dump when calling Transformation zexcel_tr_sheet
The Transformation zexcel_tr_sheet dumps when called from Class zcl_excel_writer_huge_file.

The problem is that this line refers to **HIGHT**: https://github.com/sapmentors/abap2xlsx/blob/68707e1469e8a39982f0a672fb8788c3be15337f/src/zexcel_tr_sheet.xslt.source.xml#L116
but the corresponding line of `zcl_excel_writer_huge_file` refers to **HEIGHT** : https://github.com/sapmentors/abap2xlsx/blob/68707e1469e8a39982f0a672fb8788c3be15337f/src/zcl_excel_writer_huge_file.clas.abap#L163

sandraros:
Fix #553 and also "fit to page" which didn't work at all.

### DIFF
--- a/src/zcl_excel_writer_huge_file.clas.abap
+++ b/src/zcl_excel_writer_huge_file.clas.abap
@@ -249,7 +249,6 @@ CLASS zcl_excel_writer_huge_file IMPLEMENTATION.
           draft              TYPE c,
           errors             TYPE string,
           firstpagenumber    TYPE i,
-          fittopage          TYPE c,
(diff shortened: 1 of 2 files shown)


AndreaBorgia-Abo (approved):
Verified transformation before and after, change makes sense.
The test program demonostrates the issue before merging and works fine afterwards, LGTM.

PR #918, opened 2021-12-14
Fix enhancement categories of several structures

christianguenter2:
Fixes these syntax check warnings
grafik


sandraros:
That will handle the messages "After being enhanced, the structure "&A" might contain non-character-like components. It would then no longer be..." indicated in #874.

Solution in the PR is to change the enhancement category of concerned DDIC structures from "Can be enhanced (deep)" to "Can be enhanced (character-like)".

sandraros (approved):
LGTM 👍

PR #919, opened 2021-12-14
Fix links after org. move

AndreaBorgia-Abo:
grep for the win: hopefully this is the end of it.


AndreaBorgia-Abo (commented):
changes are as expected

larshp (approved):

PR #927, opened 2021-12-24
fix typo: hasehd -> hashed

AndreaBorgia-Abo:
No functional changes, fixed a typo to have consistent naming


AndreaBorgia-Abo (commented):
changes are as expected

sandraros:
I'm afraid of changing any public member, but I think it's acceptable here because this type is only used privately. Would you mind making all these types private too? (I couldn't suggest the changes, but it'll be [feasible soon](https://github.com/github/roadmap/issues/347)) Thanks!

AndreaBorgia-Abo (commented):
types are now private

sandraros (approved):

PR #928, opened 2021-12-24
Demo WebDynpro: ICF, button text, MIME type

### LINKED ISSUE #923: abapgit does not create ICF node of webdynpro application and downloaded filename wrong
After installing ABAP2XLSX **ZDEMO_EXCEL_WDA01** does not work
After much debugging found WebDynpro needs a ICF Node /sap/bc/webdynpro/sap/zdemo_excel_wda01
**which abapgit  does not create**

I deleted and re-created the application
Ran successfully - downloaded file was wda01.txt not wda01.xls as coded in **onactionbtn_download**
Please fix this

sandraros:
Fix #923


src/demos/npl0018        89d0c0ce8687b56786306d6dd.sicf.xml
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<abapGit version="v1.0.0" serializer="LCL_OBJECT_SICF" serializer_version="v1.0.0">
+ <asx:abap xmlns:asx="http://www.sap.com/abapxml" version="1.0">
+  <asx:values>
+   <URL>/sap/bc/webdynpro/sap/NPL0018/</URL>
+   <ICFSERVICE>
+    <ICF_NAME>NPL0018</ICF_NAME>
```
AndreaBorgia-Abo, line 7: What happens if NPL0018 is already taken, definition is overwritten or pull fails? (my ONE test says: overwritten, not cool)
Are we creating this definition everywhere or simply in the NPL demo system? (yes, the service definition will be created anywhere...)
I'm not too comfortable with this, honestly.

AndreaBorgia-Abo (commented):
Ok, it works, that much I can confirm.
When it comes to webdynpro, I have close to no experience, actually none at all when it comes to the service definition!
Fingers crossed.

src/demos/npl0018        89d0c0ce8687b56786306d6dd.sicf.xml
sandraros: I did a few tests. It doesn't seem to work well if there's an existing ICF service with code "`NPL0018`" before the abapGit pull. abapGit should better create a new free number, but for now it doesn't. I guess we should open an issue at abapGit. There is also [abapgit#5064](https://github.com/abapGit/abapGit/issues/5064).

AndreaBorgia-Abo (approved):
Ok, I subscribed to [abapgit#5064](https://github.com/abapGit/abapGit/issues/5064) and I agree perhaps a new issue could be created (feel free to tag me if you do, too tired to do it now)
For the moment, hopefully the NPL prefix will prevent immediate clashes, so I'd say we merge this and see how it fares.

Issue #524, opened 2018-03-08
Reduction of code footprint


sigitas-stumbras:
This is more of a feature suggestion to reduce the code footprint when using abap2xlsx, not an actual issue.
1. lo_style_boxed = lo_excel->add_new_style( **io_clone_of = lo_style_base** ).
2. lo_excel_worksheet->**set_cell( )** to have an optional **io_style** paramenter, to pass reference to zcl_excel_style directly instead of requiring additional variable top pass guid to ip_style parameter.
3. zcl_excel_worksheet->**set_cell( )** and **change_cell_style( )** to have an optional **ip_cell** paramater, that would accept Excel style cell adress. Examples:
lo_excel_worksheet->set_cell( **ip_cell = 'B1'** io_style = lo_style ip_value = 'value' ).
lo_excel_worksheet->set_cell( **ip_cell = 'C1:E3'** io_style = lo_style ip_value = 'value' ). - Merger and all affected cells are handled automatically.
**ip_row_to** and **ip_column_to** parameters could also be included in both methods as an alternative capability to specify a range of cells

I could do the programing, but I would like to hear some feedback first. I really find this project quite amazing and usefull and I would like to contribute.

ivanfemia:
Hi,

1. I like the idea could be useful
1. maybe we can consider to remove the guid functionality at all (in the future? to guarantee backward compatibility). we may need to do an impact analysis on this.
1. Cool functionality

Thank you for your availability to contribute!
Ivan

sigitas-stumbras:
Hi,

I had some second thoughts considering the second point in my suggestions.

Although adding additional optional io_style parameter is a trivial matter, I suspect you may be concerned over not overextending the interface of the method. Especially considering this in long term backward compatibility.

Ok. Here is an alternative suggestion - how about making ip_style a generic type parameter, that would accept both guid and object reference. So the same parameter could be handled with some dynamic programing, without adding additional parameter and remaining fully backward-compatible.

Sigitas

A side note: Please be open and straightforward, if you have any comments or I am doing something not quite right github-wise, contribution-wise or otherwise. I have never done this before, but I'd really like to get some first hand experience with this. Especialy - to get the hang of how these kind of projects get managed and project decisions being made.

sigitas-stumbras:
Hello again,

Some second thoughts regarding point 3.
[...]

PR #930, opened 2021-12-25
ip_columnrow in SET_CELL and other methods

sandraros:
Permits code footprint reduction as described in #524


src/demos/zdemo_excel40.prog.abap
```diff
@@ -31,10 +31,7 @@ START-OF-SELECTION.
 *    These marked cells will be used for repeatable rows/columns on printpages
 *--------------------------------------------------------------------*
     lo_worksheet->set_area(
-          ip_column_start = 1
-          ip_column_end   = 20
-          ip_row          = 1
-          ip_row_to       = 100
+          ip_range        = 'A1:T100'
```
AndreaBorgia-Abo, line 34: modified call to set_area, but not that to change_area_style: [from the explanation](https://github.com/abap2xlsx/abap2xlsx/issues/524#issuecomment-1001204292) I would have expected at least one of the change_area_style calls to be rewritten to take advantage of ip_range.

AndreaBorgia-Abo (approved):
LGTM: unit tests are ok, plus the new checker returns the same results as before the PR.

src/demos/zdemo_excel40.prog.abap
sandraros: Let's do that. In both `change_area_style`, to push people to use `ip_range`?

sandraros: Done.

AndreaBorgia-Abo (approved):

PR #902, opened 2021-12-04
Sandraros/check demo regressions

sandraros:
Fix #894

PR #940, opened 2021-12-31
Sandraros/check demo regressions 2

### LINKED ISSUE #894: Unit Tests to avoid regression in existing demo programs
Proposing this issue as an open discussion.

I'd like to know your opinion about adding "unit" tests for the demo programs. That would imply:
- to store the XLSX files generated by the demo programs in the Web Repository (W3MI objects)
- to extract the code from the demonstration programs into new classes which can be called from the demo programs and from a test program
- to create a test program whose ABAP Unit test checks that the file generated by each demo program 

In the future, we could also use this feature to make sure that both "Writer" and "Huge Writer" generate the same result (currently we have many features which don't work in the Huge Writer), and the same for "Reader" and "Huge Reader". This 
[program](https://github.com/sapmentors/abap2xlsx/files/7577209/abap_program_to_compare_writer_and_huge_writer.txt) can be used as a starting point for discussion on Writer/Huge Writer, it runs 2 demos online and displays the diff, which works as explained in #891, and it also contains the test classes for them.

sandraros:
Fix #894, alternative to #902.

This is a lightweight solution, the demo programs are not changed except those ones which write `sy-datum` and `sy-uzeit` (EDIT: and demo programs which read database tables to produce Excel content → fixed data provided).


AndreaBorgia-Abo:
Hmm, first thing I notice is that the checker from the older PR would run just fine with the Java GUI whereas this one has the typical UNKNOWN_DP_ERROR control issue, this time in ZEXCEL_DEMO28. I'll try again with the Windows version, then.

AndreaBorgia-Abo (approved):
Thank you for your patience with my review speed! Here are my observations:
1. not requiring an API of sorts to write a test report is arguably a pro, since users might want to adapt a test report when starting out with this library
2. we still need to account for date and time, as already noted elsewhere, but even this might not be enough: for example, in ZEXCEL_DEMO3 data coming from table SFLIGHT is different between [NPL](https://blogs.sap.com/2019/10/01/as-abap-7.52-sp04-developer-edition-concise-installation-guide/) and [A4H](https://blogs.sap.com/2021/02/15/sap-abap-platform-1909-developer-edition-available-soon/) demo systems. I don't have a suggested fix here: ATDF is not supported on all versions, rewriting the tests is perhaps not worth the effort anyway since developers are the target users here... I guess we'll simply have to document it somewhere in the program itself or via a popup
3. going through the GUI probably restricts the demo checker to the Windows version: whereas the checker from the older PR would run just fine with the Java GUI, this one has the typical UNKNOWN_DP_ERROR control issue (this time in ZEXCEL_DEMO28)
4. in A4H (for this PR) all Reader tests fail because of a spurious tag: `ht="0.0000000000000000E+00"`
5. in NPL (for the older, as reference only) many tests fail because of this extra tag: `width="9.10"` (apparently coming from CREATE_XL_SHEET but I suspect that's because of the [iterator changes](https://github.com/abap2xlsx/abap2xlsx/pull/937) which were made afterwards)
6. if the output directory already contains the test files, the checker cannot run; I think we could either create a temporary directory for each run or just allow overwriting the files, perhaps informing the user once

All in all, this is a better solution and you may un-draft this PR.
With regards to the older one, I think we could close it without merging and perhaps conserve the corresponding tree as food for thought, if anything else.

sandraros:
Thanks to you! :)
 
1. So, closing the other solution (#902) and keeping the branch for a few days, in case...
2. I'd propose to have a hidden parameter in `ZDEMO_EXCEL3` to generate hardcoded lines. I will push something.
3. Difficult to solve UNKNOWN_DP_ERROR in `ZDEMO_EXCEL28`/Java GUI.
4. I created #944. I'll propose a PR quickly.
5. The iterator fix has added `width="9.10"` for all columns and the Excel files in the Web repository should reflect that, so I wonder if you have pressed the Refresh button to call `ZDEMO_EXCEL`/regenerate the demo files in the output directory.
6. Why do you say "if the output directory already contains the test files, the checker cannot run"? I don't have any issue.

AndreaBorgia-Abo:
2. not sure if this is the only report affected but we'll soon find out :)
3. I know, but this solution is so much better that I'm willing to live with it
5. regeneration should not be needed: I wasn't jumping from one branch to the other, rather I had pulled the older PR into NPL and the newer in A4H so comparing the behaviour was a simple matter of switching window
6. because I misinterpreted the error messsage, that's why: it only happens if you leave Excel open on any of the test files, in which case it will just dump

AndreaBorgia-Abo (commented):
Almost there but not quite... and the test results are weird!

In A4H only:
* ZDEMO_EXCEL3
* ZDEMO_EXCEL10
* ZDEMO_EXCEL22 (also: top 3 rows are frozen, instead of just one, perhaps something was removed above)
* ZDEMO_EXCEL33

Both in NPL and in A4H:
* ZDEMO_EXCEL15, all reader tests (but that is #944 so no worries)

This time I pulled the same branch in both systems, of course. Tomorrow I'll look into the A4H issues, unless you already have an idea.

sandraros:
In A4H only: all these demo programs get data from tables SFLIGHT or T005T. Let me provide same solution as I did for `ZDEMO_EXCEL3` (static data not from tables).

EDIT: fixed.

AndreaBorgia-Abo:
Adding suggestion here, as already noted it's not possible to suggest changes elsewhere:
```
columns->get_column( 'COMPARE_XLSX_JUST_NOW' )->set_technical( ).
columns->get_column( 'COMPARE_XLSX_REFERENCE' )->set_technical( ).
```

sandraros:
> Adding suggestion here, as already noted it's not possible to suggest changes elsewhere

It's brand new program, all lines are new, are you sure it isn't possible to suggest changes?

src/demos/zdemo_excel_checker.prog.abap
```diff
@@ -0,0 +1,1382 @@
...
+    columns->get_column( 'WRITE_SMW0' )->set_output_length( 5 ).
+    columns->get_column( 'WRITE_SMW0' )->set_alignment( if_salv_c_alignment=>centered ).
+    columns->get_column( 'PROGRAM' )->set_output_length( 15 ).
+    columns->get_column( 'PROG_TEXT' )->set_output_length( 30 ).
+    columns->get_column( 'OBJID' )->set_output_length( 20 ).
+    columns->get_column( 'OBJ_TEXT' )->set_output_length( 50 ).
+    columns->get_column( 'FILENAME' )->set_output_length( 50 ).
+    columns->get_column( 'XLSX_JUST_NOW' )->set_technical( ).
+    columns->get_column( 'XLSX_REFERENCE' )->set_technical( ).
+
```
AndreaBorgia-Abo, line 1260:
```suggestion
columns->get_column( 'COMPARE_XLSX_JUST_NOW' )->set_technical( ).
columns->get_column( 'COMPARE_XLSX_REFERENCE' )->set_technical( ).

```

AndreaBorgia-Abo (approved):
> It's brand new program, all lines are new, are you sure it isn't possible to suggest changes?

I did not see it in the changes, perhaps because it was older. Found it now and added, thanks. Oh, another suggestion: a call transaction SMW0 to download the existing reference file, a companion function to the save button.

Anyhow, there are still some differences reported in A4H:
* ZEXCEL_DEMO3: sort order of the second sheet (Data validation) is different, but all values are present
* ZEXCEL_DEMO33: as far as I can tell the visible content is unchanged but internally some nodes still have different numbering(?)

I think we could merge this and then later understand where those differences come from.

sheet33-postPR-1

sheet33-postPR-2

sheet33-postPR-3

sandraros:
I understand, it's the column headings from SFLIGHT/T005T data elements, they may differ in each SAP version, we should fix them also.

sandraros:
Agreed to merge now if you wish. I'll work on the fix this evening.

src/demos/zdemo_excel_checker.prog.abap
AndreaBorgia-Abo: Sheesh, indentation is wrong ;(

AndreaBorgia-Abo:
> Agreed to merge now if you wish. I'll work on the fix this evening.

Regarding the remaining differences, would you like a single ticket for both NPL and A4H or would you rather keep them distinct?

sandraros:
Not sure what you mean. I think I'll do one PR (didn't have free time yesterday) to make NPL and A4H produce the same result. In `ZDEMO_EXCEL3` I forgot to fix table data instead of `SELECT * FROM SCARR`, and in `ZDEMO_EXCEL33` I think we need to fix the line of column headings with `set_cell_value( ... )` right after calling the `convert` method.

PR #957, opened 2022-01-08
release7.15.0

sandraros:


src/zcl_excel.clas.abap
```diff
@@ -156,7 +156,7 @@ CLASS zcl_excel DEFINITION
     DATA worksheets TYPE REF TO zcl_excel_worksheets .
   PRIVATE SECTION.
 
-    CONSTANTS version TYPE c LENGTH 10 VALUE '7.14.0'.      "#EC NOTEXT
+    CONSTANTS version TYPE c LENGTH 10 VALUE '7.15.0'.      "#EC NOTEXT
```
AndreaBorgia-Abo: if we move this definition to the public section, a move that would make sense in its own right for such a field, then...

src/zcl_excel_apack.clas.abap
```diff
@@ -23,7 +23,7 @@ CLASS zcl_excel_apack IMPLEMENTATION.
 
     descriptor-group_id        = 'github.com/abap2xlsx/abap2xlsx'.
     descriptor-artifact_id     = 'abap2xlsx'.
-    descriptor-version         = '7.14.0'.
+    descriptor-version         = '7.15.0'.
```
AndreaBorgia-Abo:
```suggestion
    descriptor-version         = zcl_excel=>version.
```

AndreaBorgia-Abo: ...here we could reference and always be ok.

AndreaBorgia-Abo (approved):
LGTM.
One quick, process-related confirmation: are we using the GH generated changelog as basis or is the changelog creation entirely manual?

sandraros:
Currently, both systems exist and are maintained separately. I update `changelog.txt` manually from the list of latest commits. One is for seeing the changes from within ZABAPGIT (APACK), one is for seeing it directly on GitHub via Release management.

AndreaBorgia-Abo (approved):

AndreaBorgia-Abo:
To be clear, I intentionally did not merge right away because I saw that multiple reviewers were involved and I'm not sure if you wanted feedback from all of us. Feel free to merge if one approval is enough.

sandraros:
Thanks. We didn't agree on a policy for the releases, so I think we can merge. I just propose that we create a release from time to time, every 1 or 2 months for instance, to contain enough changes, but not too much.

PR #959, opened 2022-01-08
GET_TABLE skip bottom empty rows

### LINKED ISSUE #787: Highest row calculated incorrectly
Hi, 

the method `ZCL_EXCEL_WORKSHEET->get_table()` returns rows in the exporting parameter `et_table` which are empty in the input Excel worksheet except from formatting. 

In the screenshot you can see that the excel file contains only one row with actual data, but et_table returns 9 rows (the first line is skipped according to parameter` iv_skipped_rows = 1`).

Bug

Especially in connection with the issue https://github.com/sapmentors/abap2xlsx/issues/703 regarding wrong initial date values (18991231 instead of empty) this is a problem, because empty rows are not initial and can therefore not be deleted easily from the internal table.

I saw in the debugger that the attribute `LOWER_CELL-CELL_ROW` of the class `ZCL_EXCEL_WORKSHEET` was set incorrectly (actual value 10, expected 2).

Bug3

This is an expample file with formatted but empty cells:

[Empty2.xlsx](https://github.com/sapmentors/abap2xlsx/files/6930915/Empty2.xlsx)

Thanks!

Kind regards
Bernhard

sandraros:
Fix #787


AndreaBorgia-Abo (approved):
LGTM

PR #961, opened 2022-01-08
SET_CELL_FORMULA sets formula in wrong cell

### LINKED ISSUE #932: SET_CELL_FORMULA ignores requested column
In the following program, `SET_CELL_FORMULA` sets the formula in cell `A4` instead of cell `C4` as requested:
```abap
DATA: gc_save_file_name TYPE string VALUE 'test.xlsx'.
START-OF-SELECTION.
  lo_worksheet->set_cell_formula( ip_column = 'C' ip_row = 4 ip_formula = '"te"&"xt"' ).
```

Actual result:
image

Expected result:
image

Workaround: use `SET_CELL` with `IP_FORMULA` instead of `SET_CELL_FORMULA`.

NB: if you change the code to also set a value in cell `A4` the formula goes into cell `B4`, and if you set a value in cell `B4` the formula goes into cell `A4`, so I think that Excel sets the formula in the first unused column in the row.

Concerned code in method `SET_CELL_FORMULA` (proposed solution is to initialize the component `CELL_COORDS` after components `CELL_ROW` and `CELL_COLUMN`; the `CELL_COORDS` initialization logic already exists twice in the class so it would be interesting to have reusable code in new method `ZCL_EXCEL_COMMON=>CONVERT_COLUMN_A_ROW2COLUMNROW`, to be the opposite of the existing method `CONVERT_COLUMNROW2COLUMN_A_ROW`):
https://github.com/abap2xlsx/abap2xlsx/blob/26a42d8bdb50480595d9e6693e0b79eb4a044c90/src/zcl_excel_worksheet.clas.abap#L3529-L3536

sandraros:
Fix #932


AndreaBorgia-Abo (approved):
Moving the coordinate conversion to a method with test is good, I like that.
Demo program shows error before but not after PR, ok.
LGTM.

PR #970, opened 2022-01-15
Table with column headings of more than 40 characters

sandraros:
- Please first review #971 
- New field `SCRTEXT_H` "huge text" in field catalog, of type String
- `ZDEMO_EXCEL47` adapted for demonstration


sandraros:
Fix done in other PR -> #1051

PR #974, opened 2022-01-22
Add missing exception handling

aspeitkamp:
Fixes shortdump in ZDEMO_EXCEL32 and custom programs if the ALV bind functionality is being used.
The dumps are caused by the missing handling of the exception CX_SALV_BS_SC_RUNTIME_INFO.


sandraros (approved):
Thanks. LGTM 👍

PR #966, opened 2022-01-15
Create publishing-a-new-release.md

sandraros:
Procedure targeted to maintainers of the project

PR #975, opened 2022-01-22
Fix apack manifest 7.15.0

sandraros:
(not sure why it wasn't changed with #957)


AndreaBorgia-Abo (approved):
Unless there is a way to link .apack-manifest.xml to "CONSTANTS version" in src/zcl_excel_apack.clas.abap I'm afraid the only possible prevention is adding this step to the release checklist.

sandraros:
I think the issue is that abapGit doesn't actualize the version when the APACK class is already in memory. Thus need to restart abapGit to consider any change in the APACK class. I created [abapGit/#5284](https://github.com/abapGit/abapGit/issues/5284). Meanwhile I have updated the checklist in #966.

PR #982, opened 2022-01-30
SET_CELL use output length for conversion exits

### LINKED ISSUE #320: Method GET_VALUE_TYPE of class ZCL_EXCEL_WORKSHEET causes problems when converted value doesn't fit in the import field
In the demo report ZDEMO_EXCEL33 the table T005T is used to provide the sample data. It has the column SPRAS with the data type SPRAS which is only one character long. For this field the conversion exit CONVERSION_EXIT_ISOLA_OUTPUT is found in method GET_VALUE_TYPE of class ZCL_EXCEL_WORKSHEET. That causes the input value i.e. 'D' to be converted to 'DE'. But as the input value is the template for the value which holds the output of method GET_VALUE_TYPE in method SET_CELL

_First we get reference into local var.
    CREATE DATA lo_value LIKE ip_value.
    ASSIGN lo_value->_ TO <fs_value>.
    <fs_value> = ip_value.

the result in <fs_value> is again 'D'. That is no problem with German or English. But if the input language is Bulgarian where the SAP language code is 'W' then the conversion exit returns 'BG' which is stored in the Excel file as 'B'. When this value is read back from Excel to ABAP the 'B' will be interpreted as the SAP language code and will result in Hebrew.

Here you can find some SAP language codes: http://help.sap.com/saphelp_nw04/helpdata/en/c1/ae563cd2ad4f0ce10000000a11402f/content.htm

Please let's discuss how we tackle this problem.

sandraros:
Fix #320


AndreaBorgia-Abo (approved):
The default for CONV_EXIT_LENGTH is reasonable, demos are ok and the regression checker is up to date, LGTM!

PR #984, opened 2022-01-30
Reader: custom paper height and width

### LINKED ISSUE #396: Template Reader 2007 - Paper Size not read
Missing feature:  Readerclass does not read PaperSize ( A4,A5, Letter legal ) etc.

sandraros:
Fix #396


AndreaBorgia-Abo (approved):
As noted in the bugreport, this fix may not cover all versions but it is better than nothing and the changes are limited enough in scope to be acceptable anyway. So, LGTM.

AndreaBorgia-Abo (approved):
LGTM
(everybody reading this some time from now: yes, it is a partial fix, we know it!)

Issue #440, opened 2016-06-08
Autofilter with more values does not work


miskohut:
**EDIT 2024 by @sandraros: I have added a [comment below](https://github.com/abap2xlsx/abap2xlsx/issues/440#issuecomment-2156021645) to fully describe and reproduce the bug.**

Hi, 

I have defined autofilter for 13th row. And I want to set more values to first column. I do it by:

```

lv_cell_value = 'Sum'.

  lo_autofilter->set_value(
    exporting
      i_column = 1    " Cell Column
      i_value  = lv_cell_value
  ).

  lv_cell_value = 'ProjectA'.

  lo_autofilter->set_value(
    exporting
      i_column = 1    " Cell Column
      i_value  = lv_cell_value
  ).

  lv_cell_value = 'ProjectB'.

  lo_autofilter->set_value(
    exporting
      i_column = 1    " Cell Column
      i_value  = lv_cell_value    " Cell Column
  ).

  lv_cell_value = 'ProjectC'.

  lo_autofilter->set_value(
    exporting
      i_column = 1    " Cell Column
      i_value  = lv_cell_value   " Cell Column
  ).
```

When excel is generated, the values in filter are selected, but they are not displayed. I have to open the filter, the values are checked but I have to press 'OK' in order to 'apply' the filter.

I'm doing this because I want all values to be selected when excel is generated. But when I just set filter without setting values then nothing is selected in the filter.

I have added one value to report `ZDEMO_EXCEL33` and the result is the same. All values defined in code are checked in the filter but you must press 'OK' to apply the filter. Could you please help me with that? Thank you in advance.

gregorwolf:
Any chance that you fork the project, create a fix and then provide a pull request?

miskohut:
I'm not sure if I understood the question. Are you asking me if I fixed the problem myself? If you do then no, I was trying to debug the writer class but with no success.

StefanSchmoecker:
Maybe we should merge this issue #440 with issues  #368 and #421 into one large issue "Multiple errors with autofilters"?

spid3r:
Hi guys,

the problem can be solved in Method GET_FILTER of class ZCL_EXCEL_CONVERTER_ALV.
The following coding should resolve all problems related to incorrect filtering due to multiple excluding values:

```
 TYPES: BEGIN OF ty_field_range,
           fieldname TYPE lvc_fname,
           range_tab TYPE REF TO data,
[...]

Issue #421, opened 2016-02-03
Filter issue with multiple exclude values


admiralmaggie:
When using an ALV layout with a filter which includes multiple exclude values, incorrect rows are included in auto filtered worksheet. A single exclude condition works fine. I have replicated and corrected this behavior in my local copy. The issue is related to how filter records are applied in the following method:

`ZCL_EXCEL_CONVERTER_ALV->GET_FILTER()`

ALV provides a function to dynamically apply filter to an internal table:

```
  CALL FUNCTION 'LVC_FILTER_APPLY'
    EXPORTING
      it_filter       = wt_filt
    IMPORTING
      et_filter_index = lt_filter_index
    TABLES
      it_data         = <fs_tab>.
```

Any thoughts?

PR #985, opened 2022-01-30
Filter issue with multiple exclude values

sandraros:
Work on #421 and #440, still in progress...


darnoc312:
Hello,

i want to propose a complete solution of your pull request:

1) The main problem is line 585 of zcl_excel_converter_alv in the attached code of the pull request.
   ls_filt-fieldname is still there. Must be changed to <fs_field_range>-fieldname.
   
2) I prefer to read the first line of the data table to get the component types for the first loop
   like for example method zcl_excel_converter->create_table does it too.
   Also i have an easy check if the fieldname of wt_filt exists in the data table.
   This is unfortunately gone in the attached code. if you insits on this please catch at least
   the exception COMPONENT_NOT_FOUND of method get_component_type and insert a CHECK sy-subrc = 0
   below.

3) Modus filter = abap_undefined: Using the advantage of collecting all values for each fieldname
   i delete the table line immediately if not selected. Now it is really as simple as that.

4) I included a sort for wt_filt by fieldname and a corresponding fieldname compare in the first loop
   to have no need of reading lt_field_range again. Therefore now this table is a standard table.
   
5) I included an append initial line for the range table of lt_field_range.
   That avoids the data reference for the range WA.

[zcl_excel_converter_alv-get_filter.txt](https://github.com/abap2xlsx/abap2xlsx/files/12098263/zcl_excel_converter_alv-get_filter.txt)

Do you like to continue the work on your pull request or shoud i open a new issue ?

sandraros:
Thanks, please create a new pull request. I will retire all my old draft pull requests.

sandraros:
Too old and reworked by @darnoc312, so retiring my draft PR.

PR #995, opened 2022-02-20
Workaround abapGit bug

sandraros:
The class `zcl_excel_writer_2007` cannot be pulled initially with abapGit, because it's using components from DDIC structures which are references to `zcl_excel_drawing`. When abapGit deserializes the DDIC objects, it temporarily switches the class to the generic type `OBJECT`, and when any PROG or XSLT program refers to the class `zcl_excel_writer_2007` or its subclasses, it detects that the class has syntax error for the lines below, because at the time of activation, `drawing` and other components are of generic type `OBJECT`:
- `lv_value = <fs_drawings>-drawing->get_index( ).`
- `REPLACE '#' IN lv_value WITH <fs_drawings>-drawing->get_media_name( ).`
- `ls_drawing_position = is_footer-left_image->get_position( ).`
- etc.

A little change in code, and the problem is avoided.

### DIFF
--- a/src/demos/zdemo_excel24.prog.xml
+++ b/src/demos/zdemo_excel24.prog.xml
@@ -14,7 +14,7 @@
     <item>
      <ID>R</ID>
      <ENTRY>abap2xlsx Demo: Multiple sheets with different default date formats</ENTRY>
-     <LENGTH>70</LENGTH>
+     <LENGTH>67</LENGTH>
     </item>
    </TPOOL>
   </asx:values>
--- a/src/demos/zdemo_excel5.prog.xml
+++ b/src/demos/zdemo_excel5.prog.xml
@@ -14,7 +14,7 @@
     <item>
      <ID>R</ID>
      <ENTRY>abap2xlsx Demo: Conditinal formating</ENTRY>
-     <LENGTH>37</LENGTH>
+     <LENGTH>36</LENGTH>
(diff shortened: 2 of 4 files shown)


AndreaBorgia-Abo (commented):
Something still does not seem right: I cloned my [NPL demo](https://blogs.sap.com/2019/10/01/as-abap-7.52-sp04-developer-edition-concise-installation-guide/) (where note 2859159 cannot be installed because of incompatible kernel), removed all dependencies (from my examples repo) then uninstalled abap2xlsx; starting from scratch, the master branch as expected cannot be pulled but neither can the PR branch, same error on "ZEXCEL_TR_DOCPROPS_CORE". Forcing activation, though, it seems to be fine.
EDIT: hold your horses, re-checking... nope, unfortunately this happens with the very latest abapgit too.

sandraros:
Thanks. Last week-end, I could make abap2xlsx pulled without error, it needs this current fix + two fixes to abapGit (a third one was merged already). I hope we're almost done with https://github.com/abapGit/abapGit/pull/5369. The second and last one I'm working on: https://github.com/abapGit/abapGit/pull/5371. We may wait for abapGit fixes to be merged, and continue on #995 after that.

sandraros:
No more issue after #998, so PR not needed anymore.

PR #997, opened 2022-03-06
Fix diff due to abapGit changes

sandraros:


AndreaBorgia-Abo:
Anything in particular to watch out for?

sandraros:
When I pulled abap2xlsx, abapGit (latest version) has reported Diff for these 5 objects. Could you check that it's the same for you?

AndreaBorgia-Abo:
I have not seen anything wrong, should I try with an empty container?
Of course, in that case, after new PR on the demos repo.

sandraros:
I'm not sure. I need to check again since all demo programs are moved to the repository abap2xlsx/demos.

sandraros:
No more issue. Abandoning this pull request.

PR #1008, opened 2022-03-26
remove leftover int8 declaration ( fix #1007 )

### LINKED ISSUE #1007: 7.15.0: ZCL_EXCEL_WORKSHEET, local test class, method INT8: leftover int8 declaration may break compile
Currently here: https://github.com/abap2xlsx/abap2xlsx/blob/main/src/zcl_excel_worksheet.clas.testclasses.abap#L1582

The method int8 checks if int8 is available in the system, but has a leftover(?) int8 declaration - so if the system doesn't support it, it won't even compile.

AndreaBorgia-Abo:
should fix #1007


sandraros (approved):
LGTM

PR #1010, opened 2022-03-26
allow suppression of headerline in bind_table

AndreaBorgia-Abo:
Prologue: coworker wanted to dump a table without any headerline / filters, using method IF_SALV_BS_TT_UTIL~TRANSFORM of standard class CL_SALV_BS_TT_UTIL. Best I can say about the resulting solution is that it worked but it involved overwriting the catalog, not cool.

I thought such a functionality existed already in abap2xlsx and.... I was wrong, so I naturally had to implement it.
There is a [companion PR](https://github.com/abap2xlsx/demos/pull/2) for the relevant demo.

### DIFF
--- a/src/zcl_excel_worksheet.clas.abap
+++ b/src/zcl_excel_worksheet.clas.abap
@@ -170,6 +170,7 @@ CLASS zcl_excel_worksheet DEFINITION
         !ip_table               TYPE STANDARD TABLE
         !it_field_catalog       TYPE zexcel_t_fieldcatalog OPTIONAL
         !is_table_settings      TYPE zexcel_s_table_settings OPTIONAL
+        !is_show_headerline     TYPE abap_bool DEFAULT abap_true
         VALUE(iv_default_descr) TYPE c OPTIONAL
         !iv_no_line_if_empty    TYPE abap_bool DEFAULT abap_false
         !ip_conv_exit_length    TYPE abap_bool DEFAULT abap_false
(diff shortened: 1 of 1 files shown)


sandraros:
BIND_TABLE creates a Table in Excel, so the heading line is mandatory, the user can't delete the header line, she can just "Convert to Range" and then delete the header line. I don't feel comfortable with creating a Table without the header line, it should not be possible. Here is the result of `zdemo_excel10` before your fix, it's a Table, not just values:
image

I don't understand what is the exact scenario. How abap2xlsx, ALV and CL_SALV_BS_TT_UTIL are connected in your coworker scenario?

AndreaBorgia-Abo:
Business requirement was to dump an internal table to XLSX _without_  header line, come to think of it it's not even the first time I hear this request. It must be a scenario where the format is agreed upon in advance so the header line is just a nuisance.

Since abap2xlsx was not present in the system (don't get me started...) I suggested [an alternative.](https://github.com/AndreaBorgia-Abo/ABAP-Examples/blob/main/src/forum/zabo_itab_to_excel.clas.abap) which unfortunately had the same issue, so they resorted to a less than ideal workaround. There is no technical connection: just another option we tried before I decided to take a stab at implementing it.

Back to the bind_table method: I don't know what Excel does with that menu but I know what we do in this method and that is simply a column by column scan of an internal table, with a set_cell call for each value.  Libreoffice has no issues opening such a file, generated with the modified demo and the blank flag. Did you run into trouble with Excel? Which version? I'm going to try with 2019 and let you know.

AndreaBorgia-Abo:
It seems Libreoffice is far more forgving in this specific case than Excel 2019, which complained about circular references while opening the file (LO would be happy simply displaying them as broken) and left the background color in place for the first line, not cool.

More to the point: **this PR is completeley un-necessary** so I will close it now; what my colleague wanted is already available with method set_table!
set_table-abap
set_table-result

Thanks for the reality-check, @sandraros :)

sandraros:
OK nice, thank you for the checking!

PR #1017, opened 2022-04-17
enable row filters when binding alv

### LINKED ISSUE #807: Method BIND_ALV creates Excel file which ignores the ALV row filter

AndreaBorgia-Abo:
Preliminary work, for feedback:
- added i_table = abap_false in BIND_ALV (ZDEMO_EXCEL33)
- with table false, setting "filter" option in method BIND_ALV enables filter detection later; this happens in GET_FILTER (ZCL_EXCEL_CONVERTER_ALV) which takes the filter, applies it and returns table "et_filter" with the row numbers of the filtered items
- BIND_CELLS (ZCL_EXCEL_CONVERTER) and LOOP_NORMAL (ZCL_EXCEL_CONVERTER) then create an autofilter based on the values of those cells
- interesting, SET_AUTOFILTER_AREA (ZCL_EXCEL_CONVERTER) sets a weird filter area from R1C1 to R0C4, seems to be fine though.

On the left, the file generated with the simple patch of this PR opening the file in Excel 2019, I can see an autofilter is set for a specific value (42692) in column D (flight date) but does not take effect because of the formatting.
On the right the same autofilter once the file has been re-saved from Excel, after setting the filter for year 2016.
immagine
I repeated the test with column S (agency number), same thing: autofilter created but not working because of formatting.

Clearly still some work to do, I'm documenting it here in case anyone has some tips, until I have some more time to dig into this.


AndreaBorgia-Abo:
Removing the milestone for 7.16 since I don't know when I'll be able to work on this and I'd like to release the new version.

PR #1044, opened 2022-07-11
endless loop in bind_table

### LINKED ISSUE #1043: Excel file creation takes extremely long time
Hi,

my intention is to schedule the Controlling period closing reports via background jobs and send the results as attachements in e-mails.

The excel creation for small result table works fine. But with larger results (5757 rows and 57-60 columns) i have massive problems. I have already tried to use ZCL_EXCEL_WRITER_2007 and also ZCL_EXCEL_WRITER_HUGE_FILE. I cancelled both background jobs after 60.000 seconds without any result. 

When i run the report online the extraction ALV GRID to Excel only takes a view seconds. 

This is my coding (see also URL: https://blogs.sap.com/2021/07/28/sending-spool-list-into-excel-format-through-email-attachment/, comment from Sandra Rossi July 28, 2021 at 5:49 pm)
```abap
"=================`
" EXTRACT ALV DATA
"=================
types TY_RANGE_AGR_NAME type range of AGR_NAME.
field-symbols <TABLE> type TABLE.

CL_SALV_BS_RUNTIME_INFO=>SET( DISPLAY  = ABAP_FALSE METADATA = ABAP_FALSE DATA = ABAP_TRUE ).
submit  SAPKKA00BG  using selection-set 'TEST' exporting list to memory
  and return.
CL_SALV_BS_RUNTIME_INFO=>GET_DATA_REF( importing R_DATA = data(LR_DATA) ).
assign LR_DATA->* to <TABLE>.


  "==================
  " CREATE EXCEL FILE
  "==================
  data(WORKBOOK) = new ZCL_EXCEL( ).
  data(WORKSHEET) = WORKBOOK->GET_ACTIVE_WORKSHEET( ).
  WORKSHEET->BIND_TABLE( IP_TABLE = <TABLE> ).
*  data(XLSX_GENERATOR) = cast ZIF_EXCEL_WRITER( new ZCL_EXCEL_WRITER_2007( ) ).
  data(XLSX_GENERATOR) = cast ZIF_EXCEL_WRITER( new ZCL_EXCEL_WRITER_HUGE_FILE( ) ).
  data(XLSX_DATA) = XLSX_GENERATOR->WRITE_FILE( WORKBOOK ).


  "=================
  " SEND EMAIL
  "=================
  data(EMAIL) = CL_DOCUMENT_BCS=>CREATE_DOCUMENT(
                    I_SUBJECT = 'CO Period-Closing-File' I_TYPE = 'TXT'
                    I_TEXT = value #( ( LINE = |See attachment| ) ) ).
  EMAIL->ADD_ATTACHMENT( I_ATTACHMENT_TYPE = 'XLS' I_ATTACHMENT_SUBJECT = 'Attachment'
                    I_ATTACHMENT_SIZE = |{ XSTRLEN( XLSX_DATA ) }|
                    I_ATT_CONTENT_HEX = CL_BCS_CONVERT=>XSTRING_TO_SOLIX( XLSX_DATA )
                    I_ATTACHMENT_HEADER = value #( ( LINE = |&SO_FILENAME=TEST.xlsx| ) ) ).

  data(BCS) = CL_BCS=>CREATE_PERSISTENT( ).
  BCS->SET_DOCUMENT( EMAIL ).
  BCS->ADD_RECIPIENT( I_RECIPIENT = CL_CAM_ADDRESS_BCS=>CREATE_INTERNET_ADDRESS( 'max.mustermann@mustermail.com' ) ).
  BCS->SET_SEND_IMMEDIATELY( ABAP_TRUE ). " prefer avoiding that, batch it instead
  BCS->SEND( ).
  commit work.
```
The attached file is an export from the result of report: SAPKKA00BG
[EXPORT.XLSX](https://github.com/abap2xlsx/abap2xlsx/files/9054558/EXPORT.XLSX)

Do you have any advice for me?

Thank you very much.

Best regards, 

Marc

sandraros:
Fix #1043

### DIFF
--- a/src/zcl_excel_worksheet.clas.abap
+++ b/src/zcl_excel_worksheet.clas.abap
@@ -2975,7 +2975,7 @@ CLASS zcl_excel_worksheet IMPLEMENTATION.
                    <scrtxt2>          TYPE any,
                    <scrtxt3>          TYPE any.
 
-    " Due restrinction of new table object we cannot have two column with the same name
+    " Due to restrictions in new table object we cannot have two columns with the same name
     " Check if a column with the same name exists, if exists add a counter
     " If no medium description is provided we try to use small or long
 
--- a/src/zcl_excel_worksheet.clas.abap
+++ b/src/zcl_excel_worksheet.clas.abap
@@ -2965,6 +2965,9 @@ CLASS zcl_excel_worksheet IMPLEMENTATION.
     DATA: lt_field_catalog      TYPE zexcel_t_fieldcatalog,
           lv_value_lowercase    TYPE string,
           lv_scrtext_l_initial  TYPE scrtext_l,
+          lv_long_text          TYPE string,
+          lv_max_length         TYPE i,
+          lv_temp_length        TYPE i,
           lv_syindex            TYPE c LENGTH 3,
           lt_column_name_buffer TYPE SORTED TABLE OF string WITH UNIQUE KEY table_line.
     FIELD-SYMBOLS: <ls_field_catalog> TYPE zexcel_s_fieldcatalog,
@@ -3010,9 +3013,7 @@ CLASS zcl_excel_worksheet IMPLEMENTATION.
       ENDIF.
 
       lv_scrtext_l_initial = <ls_field_catalog>-scrtext_l.
-      IF strlen( lv_scrtext_l_initial ) > 38.
-        lv_scrtext_l_initial = substring( val = lv_scrtext_l_initial len = 38 ).
-      ENDIF.
+      DESCRIBE FIELD <ls_field_catalog>-scrtext_l LENGTH lv_max_length IN CHARACTER MODE.
       DO.
         lv_value_lowercase = <ls_field_catalog>-scrtext_l.
         TRANSLATE lv_value_lowercase TO LOWER CASE.
@@ -3022,7 +3023,14 @@ CLASS zcl_excel_worksheet IMPLEMENTATION.
           EXIT.
         ELSE.
           lv_syindex = sy-index.
-          CONCATENATE lv_scrtext_l_initial lv_syindex INTO <ls_field_catalog>-scrtext_l.
+          CONCATENATE lv_scrtext_l_initial lv_syindex INTO lv_long_text.
+          IF strlen( lv_long_text ) <= lv_max_length.
+            <ls_field_catalog>-scrtext_l = lv_long_text.
+          ELSE.
+            lv_temp_length = strlen( lv_scrtext_l_initial ) - 1.
+            lv_scrtext_l_initial = substring( val = lv_scrtext_l_initial len = lv_temp_length ).
+            CONCATENATE lv_scrtext_l_initial lv_syindex INTO <ls_field_catalog>-scrtext_l.
+          ENDIF.
         ENDIF.
       ENDDO.
 
--- a/src/zcl_excel_worksheet.clas.abap
+++ b/src/zcl_excel_worksheet.clas.abap
@@ -2964,12 +2964,13 @@ CLASS zcl_excel_worksheet IMPLEMENTATION.
 
     DATA: lt_field_catalog      TYPE zexcel_t_fieldcatalog,
           lv_value_lowercase    TYPE string,
+          lv_scrtext_l_initial  TYPE scrtext_l,
           lv_syindex            TYPE c LENGTH 3,
           lt_column_name_buffer TYPE SORTED TABLE OF string WITH UNIQUE KEY table_line.
     FIELD-SYMBOLS: <ls_field_catalog> TYPE zexcel_s_fieldcatalog,
-                   <scrtxt1> TYPE any,
-                   <scrtxt2> TYPE any,
-                   <scrtxt3> TYPE any.
+                   <scrtxt1>          TYPE any,
+                   <scrtxt2>          TYPE any,
+                   <scrtxt3>          TYPE any.
 
     " Due restrinction of new table object we cannot have two column with the same name
     " Check if a column with the same name exists, if exists add a counter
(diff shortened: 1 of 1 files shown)


AndreaBorgia-Abo:
On it! While I test this, fancy looking at that date/time format PR?

src/zcl_excel_worksheet.clas.abap
```diff
@@ -2964,12 +2964,13 @@ CLASS zcl_excel_worksheet IMPLEMENTATION.
...
+          lv_scrtext_l_initial  TYPE scrtext_l,
           lv_syindex            TYPE c LENGTH 3,
           lt_column_name_buffer TYPE SORTED TABLE OF string WITH UNIQUE KEY table_line.
     FIELD-SYMBOLS: <ls_field_catalog> TYPE zexcel_s_fieldcatalog,
-                   <scrtxt1> TYPE any,
-                   <scrtxt2> TYPE any,
-                   <scrtxt3> TYPE any.
+                   <scrtxt1>          TYPE any,
+                   <scrtxt2>          TYPE any,
+                   <scrtxt3>          TYPE any.
```
AndreaBorgia-Abo, line 2976:
```suggestion
                   <scrtxt1> TYPE any,
                   <scrtxt2> TYPE any,
                   <scrtxt3> TYPE any.
```

AndreaBorgia-Abo: Most likely a pretty-printer misfire

src/zcl_excel_worksheet.clas.abap
```diff
@@ -3008,6 +3009,10 @@ CLASS zcl_excel_worksheet IMPLEMENTATION.
         <ls_field_catalog>-scrtext_l = 'Column'.  " default value as Excel does
       ENDIF.
 
+      lv_scrtext_l_initial = <ls_field_catalog>-scrtext_l.
+      IF strlen( lv_scrtext_l_initial ) > 38.
+        lv_scrtext_l_initial = substring( val = lv_scrtext_l_initial len = 38 ).
```
AndreaBorgia-Abo: It works, with the test program as supplied.

I tweaked it [a little bit ](https://github.com/AndreaBorgia-Abo/ABAP-Examples/blob/main/src/github/zabo_abap2xlsx_bug_1043.prog.abap)and managed to get weird labels, with an asterisk in them: more than 100 fields and 38 as cutoff value does not work well. To get back to a sane behaviour, I had to lower the cutoff to 37.

I'd say we need to find the lowest between the maximum number of fields in an internal table for SAP (1024?) and the maximum number of columns in Excel (16k?), see how many chars this number takes and then work backwards from the max length of the catalog description... TLDR: 35 is probably fine.
Another option is to count the number of entries in the catalog and adjust the cutoff value as needed but, and I say so myself, this is way overkill although likely foolproof.

AndreaBorgia-Abo (changes requested):

src/zcl_excel_worksheet.clas.abap
sandraros: I don't get why you want to unalign the TYPE keyword. Current alignment is OK for me:
```abap
    FIELD-SYMBOLS: <ls_field_catalog> TYPE zexcel_s_fieldcatalog,
                   <scrtxt1>          TYPE any,
                   ...
```

src/zcl_excel_worksheet.clas.abap
sandraros: My assumption was that it's unlikely to have 100 components with the same text, so that to keep the code simple.

If we go for a fix without bug, I'd prefer a nice suffix logic to not cut characters uselessly. I couldn't think of a simple way to right align a suffix to a field without hardcoding the length. If the final code to add the suffix is a little bit complex, maybe we can simply create a separate private method to add a suffix of any length to a text of any length.

sandraros: Proposing code to handle more than 100 same texts. Hope it works.

src/zcl_excel_worksheet.clas.abap
AndreaBorgia-Abo: Never mind, my mistake: I assumed it was ok before and, also, tried to undo it which is not really obvious with GH suggestions.

src/zcl_excel_worksheet.clas.abap
```diff
@@ -2964,12 +2964,16 @@ CLASS zcl_excel_worksheet IMPLEMENTATION.
...
           lt_column_name_buffer TYPE SORTED TABLE OF string WITH UNIQUE KEY table_line.
     FIELD-SYMBOLS: <ls_field_catalog> TYPE zexcel_s_fieldcatalog,
-                   <scrtxt1> TYPE any,
-                   <scrtxt2> TYPE any,
-                   <scrtxt3> TYPE any.
+                   <scrtxt1>          TYPE any,
+                   <scrtxt2>          TYPE any,
+                   <scrtxt3>          TYPE any.
 
     " Due restrinction of new table object we cannot have two column with the same name
```
AndreaBorgia-Abo:
```suggestion
    " Due to restrictions in new table object we cannot have two columns with the same name
```

src/zcl_excel_worksheet.clas.abap
AndreaBorgia-Abo: You went for the overkill suggestion, I like that :)

AndreaBorgia-Abo (approved):
Super, it works nicely and looks readable!
Here are the relevant transitions:
```
Text Text Text Text Text Text Text Tex 9
Text Text Text Text Text Text Text Tex10
Text Text Text Text Text Text Text Tex99
Text Text Text Text Text Text Text Te100
```
LGTM!

sandraros:
@AndreaBorgia-Abo Thanks a lot for suggestions and fast review!

PR #869, opened 2021-11-02
SET_AREA/SET_AREA_FORMULA: set value or formula for whole area instead top/left only

sandraros:
Fix #867

PR #1050, opened 2022-07-22
New format dates for DD/MM/YYYY and DD/MM/YY

juancarlosrodriguezf:
Fixed format date for DD/MM/YYYY and added a new one for DD/MM/YY


AndreaBorgia-Abo (commented):
@juancarlosrodriguezf welcome and thanks for the PR.
@sandraros I agree that c_format_date_ddmmyyyy should really be as proposed but, potentially, it is a breaking change, for those programs already using this constant. What's your opinion? I'm for warning about it but merging.

sandraros:
@AndreaBorgia-Abo I agree with you that anything that changes the behavior of existing programs should be avoided. For that, the solution is to merge the PR to build a major version (7.15.0 → 8.0.0)

But I guess that merging this PR is not urgent, as there's an easy workaround to simply hardcode the value in custom programs.

Maybe we could add a label `8.0.0` to postpone the merge, or create a branch `8.0.0` and merge it only there? We might wait for other major fixes before releasing 8.0.0. We might seek in the repository of issues and past fixes for candidates of 8.0.0 (e.g. changing the default value after #869). I guess that if we add a new release, there should be a real major feature in it. So that to have a unique repository of `8.0.0` candidates, we could use the repository of issues with tag `8.0.0` (so we'd need to create an issue corresponding to #1050). Maybe we could start a separate discussion to see if there's a good reason to build a new release?

NB: if we merge this PR, we should also update either [zdemo_excel22](https://github.com/abap2xlsx/demos/blob/2f89e63babc3590ea44773873b7b78db549f4c7b/src/zdemo_excel22.prog.abap) or its [Excel file](https://github.com/abap2xlsx/demos/blob/2f89e63babc3590ea44773873b7b78db549f4c7b/src/zdemo_excel22.w3mi.data.xlsx).

AndreaBorgia-Abo:
Just a quick comment: I'm a liiiiiiitle busy right now and it's gonna be worse in the coming months, but I'll try to look at this issue during the holidays and plan ahead for the release. Ideally, I'd like to get [7.16](https://github.com/abap2xlsx/abap2xlsx/milestone/8) out of the door and upgrade our internal installation before moving on.
EDIT: the outlook of my spare time remains so-so for the coming monhts, don't hold your collective breath.

sandraros (approved):
Let's create a new constant to keep the ascending compatibility

PR #1053, opened 2022-08-13
Release 7.16.0

AndreaBorgia-Abo:
The 7.16.0 tag shows up in my repo for the release branch, I don't know how to check if it is included in the PR.

### DIFF
--- a/docs/contributing/publishing-a-new-release.md
+++ b/docs/contributing/publishing-a-new-release.md
@@ -9,7 +9,7 @@ Version numbering is based on [Semantic Versioning 2.0.0](https://semver.org/):
 - `15`: a minor release
 - `0`: patch level (bug fixes)
 
-Create a branch for this new release, change `zcl_excel_version` to indicate the new version number and push the changes to the repository
+Create a branch for this new release, change `version` in `zcl_excel` to indicate the new version number and push the changes to the repository
 
 With GitHub Desktop (or any Git console or Git user interface), [add the tag](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/managing-commits/managing-tags) `7.15.0` to this branch.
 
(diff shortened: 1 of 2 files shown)


sandraros:
Could you [add the tag](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/managing-commits/managing-tags) via Git console (or try to use abapGit, there's the feature Tag > Create, not sure if it's the same as the GitHub way), so that we can later make it appear in the home page?
image

sandraros:
Sorry, I realize that my question was answered in your initial comment. I guess we can see using a Git client. Let's merge, we'll see.

AndreaBorgia-Abo:
Well, I guess the tag only shows in the branch where it was added and it appears that it is not carried over with a PR;
I'll add a new one in this repo now and edit the instructions after releasing.

AndreaBorgia-Abo:
also, GH recommends using "v" as first letter of release-related tags, which makes sense: I had to rename my branch from "7.16.0" to "branch-7.16.0" because git was complaining about ambiguous references since 7.16.0 could be either a tag or a branch.
I will check whether older tags can be renamed without too much breakage and then edit the guideline accordingly.

AndreaBorgia-Abo:
For future reference, this is what I did:
- [checked whether tags are annotated or lightweight](https://stackoverflow.com/a/25996877): older tags by Ivan where annotated, newer tags are lightweight
- [renamed tags](https://stackoverflow.com/a/5719854), new ones are lightweight
- navigated from old 7.14.0 and 7.15.0 release links and edited them to point to the new v7.14.0 and v7.15.0 tags, so the existing releases show up pointing to the new tags
- edited the checklist with naming suggestions

TLDR if you have a local clone of the project, run `git pull --prune --tags`

PR #1054, opened 2022-08-13
Update publishing-a-new-release.md

AndreaBorgia-Abo:
- noted that tags should be created when working from upstream repository (they do not come via PR from outside, only from local branches apparently)
- added note regarding previous tag for auto-generation of release notes


sandraros (approved):
LGTM 👍

PR #1055, opened 2022-08-15
Update guidelines for demos

### LINKED ISSUE #17: sync release tags with main library
After splitting the demos from the library, we probably need a way to inform users about the appropriate version of the demos for each release.
I propose tagging this repo with release tags (es. v7.16.0), corresponding to the latest commit before the tagging in main library. We could probably do this now only for 7.16.0 but in the future it should probably be noted in the release checklist and in the Readme.

AndreaBorgia-Abo:
closes: abap2xlsx/demos#17


AndreaBorgia-Abo (commented):
changes are as expected

sandraros (approved):
LGTM 👍

PR #1068, opened 2022-10-21
refactor zcl_excel_writer_2007 - create_xl_sheet method

Dwivedirahul44:
issue #915.
massive refactor, tool-assisted and manual.
I have tried my best to avoid adding side effects, and no apparent issue found so far. 
please review carefully and provide feedback.

PR #1069, opened 2022-10-22
zcl_excel_writer- create_xl_sheet Refactor1

Dwivedirahul44:
issue #915 
earlier PR #1068 
Fresh and First PR to delegate logic to a local class.


Dwivedirahul44:
@sandraros - fresh PR with a low incision on main code.

1. method create_xl_sheet - 

- move logic to new local class.
- commented out code 
- added call to new method
2. New local class:
-  new methods to handle subnodes creation.
 commented out the code instead of deleting it.

Dwivedirahul44:
I have tried my best to keep the diffs as low as possible. feedback welcome.

sandraros:
Nice, thanks! Last thing before my review tomorrow, I see many commented lines of code, could you remove them or is there any reason why you kept them?

Dwivedirahul44:
The old code block of the method has been commented instead of deleting after moving to local class.
I have done so in assuming that it will keep diffs in github straight forward also add convenience to know what was there earlier in one look on the same file. 
The idea was to delete them after first review as they are just one block instead of distributed  comment lines at various places.

Dwivedirahul44:
I can get rid of them if you need me to, let me know.

Dwivedirahul44:
I deleted the comments and diffs are looking good too. thanks.

sandraros:
Before the review, I have added 3 commits:
- One due to abapGit changes (XML was improved), I did a separate PR, and merged them into your branch.
- One to retain texts of constant ZCL_EXCEL_WRITER=>C_XL_PRINTERSETTINGS (were removed but constant still there)
- One for cleanup of unused variables + pretty print of new local class

sandraros (changes requested):
I used an external comparison tool to compare old code versus new code, that was straight forward, and it's globally okay for me. If you agree, you might improve this way, but is optional:
- For better future diff with previous code (just in case), we might reorder these 3 blocks so that all lines of code in the local class matches the previous code and can be compared easily by comparison tool:
  - block about `o_excel_ref->create_xl_sheet_pagebreaks` to be in a new method (I know, it's 4 lines of code) before `METHOD add_xl_sheet_drawing_subnode.`
  - block `o_excel_ref->create_xl_sheet_ignored_errors` to be in a new method (I know, it's 4 lines of code) before `METHOD add_xl_sheet_table_parts_subnd.`
  - block `METHOD add_xl_sheet_sheetdata_subnode.` to be moved above `METHOD add_xl_sheet_sheetdata_subnode.`
- Renaming the methods so that to have full words, e.g. no need to have all methods named `add_xl_sheet_..._subnode` because the class is itself dedicated to processing them (`lcl_create_xl_sheet`), e.g. `add_xl_sheet_drwng4cmt_sbnde` becomes `drawings_for_comments`, etc.
- Define `lo_element_root` as attribute initialized in method `create` once instead of repeating code in many methods?
- Define `lv_relation_id` as attribute instead of passing it as parameter of 5 methods?

Dwivedirahul44:
> Before the review, I have added 3 commits:
> 
> * One due to abapGit changes (XML was improved), I did a separate PR, and merged them into your branch.
> * One to retain texts of constant ZCL_EXCEL_WRITER=>C_XL_PRINTERSETTINGS (were removed but constant still there)
> * One for cleanup of unused variables + pretty print of new local class

Thanks a lot 👍

Dwivedirahul44:
>   * block about `o_excel_ref->create_xl_sheet_pagebreaks` to be in a new method (I know, it's 4 lines of code) before `METHOD add_xl_sheet_drawing_subnode.`
>   * block `o_excel_ref->create_xl_sheet_ignored_errors` to be in a new method (I know, it's 4 lines of code) before `METHOD add_xl_sheet_table_parts_subnd.`
 - It makes sense to have a separate method for these, I will add it. 
>   * block `METHOD add_xl_sheet_sheetdata_subnode.` to be moved above `METHOD add_xl_sheet_sheetdata_subnode.`
> * Renaming the methods so that to have full words, e.g. no need to have all methods named `add_xl_sheet_..._subnode` because the class is itself dedicated to processing them (`lcl_create_xl_sheet`), e.g. `add_xl_sheet_drwng4cmt_sbnde` becomes `drawings_for_comments`, etc.

You are right, I added the prefix prior to creating the local class and kept it. I will rename the methods and you can provide additional feedback on the naming convention post changes.

> * Define `lo_element_root` as an attribute initialized in method `create` once instead of repeating code in many methods?
agree
> * Define `lv_relation_id` as an attribute instead of passing it as parameter of 5 methods?

I think there is an opportunity to make all methods lose parameters by passing  `io_worksheet`,`iv_active`,`io_document` of `create` method to a private class attribute along with your suggestion. (I might be going overboard on it).

thanks for spending time on it and sharing your feedback, Let me know your thoughts.

sandraros:
I agree to "make all methods lose parameters [...]" and use private **instance** attributes instead. Thanks for the commit, I'll review it by end of day. You may mark "changes requested" as done.

Dwivedirahul44:
> I used an external comparison tool to compare old code versus new code, that was straight forward, and it's globally okay for me. If you agree, you might improve this way, but is optional:
> 
> * For better future diff with previous code (just in case), we might reorder these 3 blocks so that all lines of code in the local class matches the previous code and can be compared easily by comparison tool:
>   
>   * block about `o_excel_ref->create_xl_sheet_pagebreaks` to be in a new method (I know, it's 4 lines of code) before `METHOD add_xl_sheet_drawing_subnode.`
>   * block `o_excel_ref->create_xl_sheet_ignored_errors` to be in a new method (I know, it's 4 lines of code) before `METHOD add_xl_sheet_table_parts_subnd.`
>   * block `METHOD add_xl_sheet_sheetdata_subnode.` to be moved above `METHOD add_xl_sheet_sheetdata_subnode.`
> * Renaming the methods so that to have full words, e.g. no need to have all methods named `add_xl_sheet_..._subnode` because the class is itself dedicated to processing them (`lcl_create_xl_sheet`), e.g. `add_xl_sheet_drwng4cmt_sbnde` becomes `drawings_for_comments`, etc.
> * Define `lo_element_root` as attribute initialized in method `create` once instead of repeating code in many methods?
> * Define `lv_relation_id` as attribute instead of passing it as parameter of 5 methods?

thanks a lot @sandraros .
The requested changes are done. I didn't see an option to mark it as done. re requested the review.
list of changes -  
- renamed methods.
- remove parameters from private methods.
- extract local constants to class level.

Future state: 
Break the methods of local class which are larger than 100 lines?
rename constants and move them to specific constant groups or extract them to own class ?

sandraros:
Maybe I'm wrong about the option to mark it done: there's a way to set resolved or accept the proposal, but it's maybe related only to the case portions of code are marked for change or with solution proposed (that I didn't do).

sandraros (changes requested):
Almost done. The following line remains at many places and that I think are useless since it's setup in the method `create`: `o_element_root = o_document->get_root_element( ).`

Dwivedirahul44:
> Almost done. The following line remains at many places and that I think are useless since it's setup in the method `create`: `o_element_root = o_document->get_root_element( ).`

I have made the changes and done some pretty print, However, I mistakenly added the changes to the wrong file(test_class instead of local class) in the previous commit, I have reverted the changes but the commit still has the file changes as its history.

I am too novice to git/GitHub to be able to undo a commit.
If commits from not needed files are a problem then I can drop this PR and create a fresh PR with commits pertaining to only the changed files.

sandraros:
Thanks. With abapGit, you can revert to the code of a previous commit, but it will create an additional "reverse" commit, the commits are never deleted with Git. But it's not a problem. What is important is the code at the end, and currently the test classes don't appear as changed, so it's fine. 👍 (NB: when we merge the changes to the main branch, all intermediates commits will be merged into one "merge" commit)

Dwivedirahul44:
Understood, thanks.

sandraros:
I pushed a commit to remove useless lines concerning `o_element_root`.

With `zdemo_excel_checker`, I saw regressions:
- Before the PR:
  image
- After the PR:
  image
  It concerned `customWidth` and `filterMode`:
  image
  image

Could you fix it and also check if there are other situations by comparing the code before/after? (not so easy)

sandraros (changes requested):
I pushed a commit to remove useless lines concerning `o_element_root`.

With `zdemo_excel_checker`, I saw regressions:
- Before the PR:
  
- After the PR:
  
  It concerned `customWidth` and `filterMode`:
  image
  image

Could you fix it and also check if there are other alike situations by comparing the code before/after? (not so easy)

Dwivedirahul44:
@sandraros - thanks for cleaning up the var. I missed pushing the changes after reverting the mess earlier.

I will use the excel checker to test and fix what I can.

Dwivedirahul44:
there were two issues that the **excel checker** found:
1.  Custom width being set when it should not 
  - Reason: method `add_cols `had a logic  adding custom width `if lo_column_default is not bound`, 
    and `lo_column_default `was not instantiated in the method.
2. Auto filter missing:
 -  Reason:   method `add_autofilter `had a guard clause checking `if lo_autofilter is bound` but the instantiation was missing. 
I have added missing instantiation logic and the regression looks good too post changes.
image

sandraros (approved):
LGTM, thanks a lot 👍

Issue #1097, opened 2023-03-25
Converter does not work if Grid not yet displayed


hardyp:
In online mode I would like to send the email before the ALV grid is displayed in the GUI.

I have a dirty trick to make this work as follows:

```ABAP
METHOD create_spreadsheet.
    "Convert SALV object into excel
    DATA: l_ws TYPE c LENGTH 10 VALUE 'ITS'.

    "If we do this (call the convertor) in the foreground we get a dump unless we do a dirty trick
    IF sy-batch EQ abap_false.
      EXPORT l_ws = l_ws TO MEMORY ID 'WWW_ALV_ITS'.
    ENDIF.

    DATA(converter) = NEW zcl_excel_converter( ).

    TRY.
        IF mo_alv IS BOUND.
          converter->convert(
            EXPORTING
              io_alv        = mo_alv
              it_table      = it_table[]
              i_table       = abap_true "Create as Table
              i_style_table = zcl_excel_table=>builtinstyle_medium2
            CHANGING
              co_excel      = mo_excel ).
        ELSE.
          RETURN.
        ENDIF.

        FREE MEMORY ID 'WWW_ALV_ITS'.

      CATCH zcx_excel INTO DATA(exception).
        DATA(message) = exception->get_text( ).
        MESSAGE message TYPE 'I'.
    ENDTRY.

  ENDMETHOD.
```

I was wondering if my "dirty trick" could somehow be embedded inside the converter class, so it does not matter in which order the conversion and the display of the ALV object happens.

Cheersy Cheers

Paul

gregorwolf:
It's better than a dump or? Keep the pull request coming.

I hope you are aware that you can copy paste ABAP code and get it correctly formatted like this:

```ABAP
WRITE: / 'Hello World'.
```

hardyp:
Just to re-iterate - if I display the SALV first and then call the convertor online all is well. If I call the converter online before displaying the SALV then I get a dump. In background everything is fine, I can call them in either order.

It would be amazingly simple to put the EXPORT/IMPORT memory statements at the start and end of the CONVERTER method. I suppose the question I am really asking is can anyone see any downside to doing this? If not, then I will indeed raise a pull request.

Code added properly. Why does it not format the first and last line?

```ABAP
METHOD create_spreadsheet.
    "Convert SALV object into excel
    DATA: l_ws TYPE c LENGTH 10 VALUE 'ITS'.

[...]

PR #1101, opened 2023-04-19
force offline mode if not in background

bigld:
I think this will do the trick without "dirty hack"


sandraros:
Does it fix #1097?

bigld:
fixes #1097

sandraros:
Thanks. I'll try to test and merge before this week-end.

sandraros:
It's difficult to understand the new code `IF sy-batch = abap_true AND io_salv->is_offline( ) = abap_false` whose result is always false (I guess).

I did these tests, the test 1 fails with ABAP 7.52 SP 0, it ignores the layout changed by the user (need to run `io_salv->get_metadata( )` to get the changed layout):
1. Display ALV, layout manually changed, then run abap2xlsx conversion
2. Prepare without displaying ALV, run abap2xlsx conversion
3. ALV data intercepted without displaying ALV, without any ALV default variant, then run abap2xlsx conversion
4. ALV data intercepted without displaying ALV, with an ALV default variant, then run abap2xlsx conversion
5. Run in background the tests 2 to 4

I have proposed another solution via the PR #1109 (waiting for abaplint fix https://github.com/abaplint/deps/pull/198 /cross check still running). It works both in ABAP 7.52 and ABAP 7.57.

Please tell me your view on that?

bigld:
perfect solution with #1109

PR #1105, opened 2023-05-06
Update colors in ALV converter + Drawing reference

### LINKED ISSUE #1102: Missing colours for ALV converter
Missing 700, 710 colours in ZCL_EXCEL_CONVERTER_ALV - CLASS_CONSTRUCTOR
My suggest:

    ls_color-col       = 7.
    ls_color-int       = 0.
    ls_color-inv       = 0.
    ls_color-fontcolor = 'FF000000'. " 000 000 000 Black
    ls_color-fillcolor = 'FFF8E5C8'. " 246 229 204
    INSERT ls_color INTO TABLE wt_colors.

    ls_color-col       = 7.
    ls_color-int       = 1.
    ls_color-inv       = 0.
    ls_color-fontcolor = 'FF000000'. " 000 000 000 Black
    ls_color-fillcolor = 'FFFDBB71'. " 255 186 106
    INSERT ls_color INTO TABLE wt_colors.

In an ideal get colours from Z customer table. Existing RGB codes(110,200...) doesn`t suit our client.

KonstantinKhrushchev:
Fix [#1102](https://github.com/abap2xlsx/abap2xlsx/issues/1102)
Optimization [#1107](https://github.com/abap2xlsx/abap2xlsx/issues/1107)


larshp:
instead of using SAP standard data elements `GUID_16` & `GUID_32` which are not used in abap2xslsx currently, and also not [released](https://abapedia.org/steampunk-2302-api/). Is it possible to use some of the already existing/used data elements?

KonstantinKhrushchev:
change guid_16 -> zexcel_guid, guid_32 -> char32

src/zcl_excel_writer_2007.clas.abap
```diff
@@ -2828,7 +2831,8 @@ CLASS zcl_excel_writer_2007 IMPLEMENTATION.
           lv_row                  TYPE string, " zexcel_cell_row.
           lv_col_offset           TYPE string,
           lv_row_offset           TYPE string,
-          lv_value                TYPE string.
+          lv_value                TYPE string,
+          lv_guid                 TYPE guid_32.
```
larshp: `GUID_32` still used here

src/zcl_excel_writer_2007.clas.abap
```diff
@@ -2604,7 +2607,7 @@ CLASS zcl_excel_writer_2007 IMPLEMENTATION.
           lo_drawing      TYPE REF TO zcl_excel_drawing.
 
     DATA: lv_value   TYPE string,
-          lv_counter TYPE i.
+          lv_guid    TYPE char32.
```
larshp: assuming this is the correct type, suggest using TYPE C instead of SAP standard CHAR32, which is currently not used in abap2xslx

```suggestion
          lv_guid    TYPE c LENGTH 32.
```

KonstantinKhrushchev:
@larshp thanks for review.
corrected

darnoc312:
Do you think this is senseful for charts too ?
IF so provide it for charts too, else let it for charts in the previous way (counter).
Currently you have a mix for charts and ZDEMO_EXCEL39 does not work anymore because of not matching rId.

For the moment i included the following in CREATE_XL_DRAWINGS_RELS recovering the field lv_counter so it works again.
lv_counter must be counted always because the while loop in CREATE_XL_DRAWINGS does too.

image

sandraros:
@darnoc312 Could you create a new issue and explain exactly what happens with `ZDEMO_EXCEL39`? (I don't get what means "does not work anymore because of not matching rId"; after that, you can propose a pull request containing your fix, text is always better than screenshot). Concerning the current issue, you can do an advance test in your system to see if the current fix works (pull the commit of the pull request). That will help us if you do a feedback on it. Thank you.

PR #1120, opened 2023-07-10
Fix ALV converter with initial layout variant

### LINKED ISSUE #1119: SALV: Download with initial layout
Hello,

in addition to ticket #745 one point to discuss.

If the initial layout is set with this statement lo_layout->set_initial_layout( 'ABCD' ) it is not handled in the method ZCL_EXCEL_CONVERTER_SALV_TABLE->LOAD_DATA

This leads to wrong output.

Here is a test report:
[issue_report.txt](https://github.com/abap2xlsx/abap2xlsx/files/11982453/issue_report.txt)

I changed this by adding ls_vari-variant   = lo_layout->get_initial_layout( ). in line 38 of the method.

What do you think about it?

thanks and regards
Matthias

oberhm:
load initial layout in converter class

Fix #1119


sandraros:
All good, thank you very much! 👍

PR #1131, opened 2023-08-04
Autofilter Solutions for excel writer

darnoc312:
Continued from pull request https://github.com/abap2xlsx/abap2xlsx/pull/985 and trying to solve https://github.com/abap2xlsx/abap2xlsx/pull/1017, but i can not reproduce any problems with date format.

Update zcl_excel_converter_alv->get_filter to avoid too many values.

Update zcl_excel_writer_2007->create_xl_sheet_sheet_data to work with multiple values for multiple columns.
Old field l_autofilter_hidden = 'X' is now represented by an entry for each filter column in table lt_cols2match.

If you use for example ZDEMO_EXCEL32 for testing be careful using ZDEMO_EXCEL33 at the same time.
Unfortunately ZDEMO_EXCEL33 calls method zcl_excel_converter->set_option instead of using the is_option parameter of the immediately following zcl_excel_converter->convert method call.
So these options, which don't contain filter = 'X', remain and the defaults of method zcl_excel_converter->init_option are not used for any further convert method call with ALV (like in ZDEMO_EXCEL32) until the 'EXCEL_username' record is deleted from database INDX.
Otherwise the only workaround is the explicit use of the is_option parameter of convert method call like attached to https://github.com/abap2xlsx/abap2xlsx/pull/1017.

### DIFF
--- a/src/not_cloud/zcl_excel_converter.clas.abap
+++ b/src/not_cloud/zcl_excel_converter.clas.abap
@@ -1698,7 +1698,7 @@ CLASS zcl_excel_converter IMPLEMENTATION.
 
 * Let's check for filter.
     IF wo_autofilter IS BOUND.
-      ls_area-row_start = 1.
+      ls_area-row_start = w_row_int.
       lt_values = wo_autofilter->get_values( ) .
       SORT lt_values BY column ASCENDING.
       DESCRIBE TABLE lt_values LINES l_lines.
--- a/src/not_cloud/zcl_excel_converter_alv.clas.abap
+++ b/src/not_cloud/zcl_excel_converter_alv.clas.abap
@@ -388,11 +388,11 @@ CLASS zcl_excel_converter_alv IMPLEMENTATION.
 
   METHOD get_filter.
     TYPES: BEGIN OF ts_field_range,
-             fieldname TYPE lvc_fname,
+             fieldname TYPE fieldname,
              range_tab TYPE REF TO data,
            END OF ts_field_range.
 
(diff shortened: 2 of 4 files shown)


darnoc312:
Update zcl_excel_autofilter because:
1) The result of zcl_excel_autofilter->get_values is expected to be sorted by for example lcl_create_xl_sheet->add_autofilter. There lv_column is used as group change field. But the mt_filters table and its values in zcl_excel_autofilter used via LOOP AT by zcl_excel_autofilter->get_values are defined as HASHED. The loops return the data in the random order of insertion and not in the order of the table key. So both tables must be changed to SORTED.
For Demonstration. The following code returns 2,1  and not 1,2:
DATA: lt_hash TYPE HASHED TABLE OF zexcel_cell_column WITH UNIQUE KEY table_line,
lv_col  LIKE LINE OF lt_hash.
lv_col = 2.
INSERT lv_col INTO TABLE lt_hash.
lv_col = 1.
INSERT lv_col INTO TABLE lt_hash.
LOOP AT lt_hash INTO lv_col.
  WRITE / lv_col.
ENDLOOP.


2) zcl_excel_converter->set_autofilter_area sets row_start, col_start and col_end but leaves row_end initial. This should be completed by the next zcl_excel_autofilter->validate_area. But since the changes of pull request #822 this is impossible. The row and column validations for less 1 and greater maximum must take place before the validations for start > end. So the new code of #822 must be placed exactly at the same place as the at that time removed code.

Issue #1134, opened 2023-08-17
CSV export - date and time fields are in excel format


oberhm:
Dear all,

we use the "CSV Writer" class to transfer a file.
In the source table we have date and time fields.

The CSV Writer exports the fields in excel format.
e.g. Date 17.08.2023 => 45155

We found a possible solution for that:
When the field lv_data_type is set in the method SET_CELL of the class ZCL_EXCEL_WORKSHEET to "D" or "T" the exporter class handles the fields correct.

If this might by a solution, I will create a pull request

thanks 
Matthias

sandraros:
Sorry for the delay. Are you sure there is a bug in the CSV writer? Could you post a minimal reproducible example so that we can check ourselves? Thank you!

oberhm:
[z_abap2xlsx_issue_1134.txt](https://github.com/abap2xlsx/abap2xlsx/files/12697525/z_abap2xlsx_issue_1134.txt)

Here is an example for the problem.
In XLSX download the date field is set with 45155, which is converted by Excel to real date 17.08.2023. In the csv export the value is also set to 45155, which is no date. If the csv is opened without excel the field is a number and no date.

I think in csv export, the date field should by a date in real format (not the excel internally format).

sandraros:
Thank you. I agree 45155 is buggy in CSV. Your program will help me analyze, and I'll confirm after week-end, based on your latest comments, probably I'll agree on your PR and merge it.

oberhm:
Hi,
after your comment yesterday I debugged again.
Maybe another solution could help, but here I'm not familiar with the data which abap2xlsx is using.

in the method CREATE_CSV of class ZCL_EXCEL_WRITER_CSV there is a coding which is called "Retrieve Cell Style format and data type"
The field cell_style is filled in the structure but this coding did not fill the variable ls_numfmt. 
Maybe we get here a correct style then the data_type is set here and must not be set in the method SET_CELL.

darnoc312:
This is my suggestion:
The NUMFMT Field is retrieved from the dynamic styles. (excel->get_styles_iterator( ) and following WHILE Loop)

At the begining of zcl_excel_writer_2007->create( ) are the following lines

* Start of insertion # issue 139 - Dateretention of cellstyles
  me->excel->add_static_styles( ).
* End of insertion # issue 139 - Dateretention of cellstyles

to create a dynamic style for any static style if not done so far.
Else some dynamic styles could be missing.

[...]

PR #1140, opened 2023-09-21
Fix CSV Export with date and time in excel format

oberhm:
fixes the issue #1134
If the csv exporter is used, the date and time fields get converted to the excel format.
This leads to problem when opening the csv file not in excel


sandraros:
Thank you for volunteering in abap2xlsx. I don't have time to look at it before middle of next week, but I feel that it's a little bit dangerous to change in the method `set_cell` of `zcl_excel_worksheet` while the issue is in `zcl_excel_writer_csv`, as it may impact all other classes and client programs. The format must internally remain like `45155` instead of `17.08.2023` (to reuse your example). I'm not sure there is a bug, let me ask you a question in the original issue (#1134).

oberhm:
Hi.
I agree with you, that these change is very dangerous.

But the variable lv_data_type is set two times in the coding before:
- if ip_data_type is supplied
- if the lv_value_type is a kind of stringlike

The lv_data_type is transferred to the sheet content field data_type

I could not see any big problem, if the field data_type is set correctly every time.
Actually the field data_type is empty in case of all data types except type stringlike.

I will response in the issue with an example.

oberhm:
This request is obsolet. see comments in #1134

sandraros (approved):
LGTM Thank you! 👍

PR #1144, opened 2023-10-03
read all header/footer (odd/even) elements

aspeitkamp:
Previously only the "oddFooter" was read and stored. Read the missing elements as well.

### DIFF
--- a/src/zcl_excel_reader_2007.clas.abap
+++ b/src/zcl_excel_reader_2007.clas.abap
@@ -2859,6 +2859,21 @@ CLASS zcl_excel_reader_2007 IMPLEMENTATION.
         ls_odd_footer-left_value = lo_ixml_hf_value_elem->get_value( ).
       ENDIF.
 
+      lo_ixml_hf_value_elem = lo_ixml_headerfooter_elem->find_from_name_ns( name = 'oddHeader' uri = namespace-main ).
+      IF lo_ixml_hf_value_elem IS NOT INITIAL.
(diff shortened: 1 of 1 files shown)


sandraros:
Thank you for proposing a fix. Could you please explain how to reproduce the issue you have fixed, in order to test before (reproduce) and after (check the fix)?

aspeitkamp:
Yes of course. 
I have attached a small sample report to be executed pre- and post the patch. It can be used to read the attached template.xlsx file, which contains a header and footer section. It will add some irrelevant data to the excel worksheet and save the result in a new excel file (result.xlsx).

Results:
If the report is run before my patch is applied, the header section will be missing in the resulting excel file.
After the patch is applied, the header section should be available in the resulting excel file.

[template.xlsx](https://github.com/abap2xlsx/abap2xlsx/files/12906684/template.xlsx)
[zexcel_template_test.txt](https://github.com/abap2xlsx/abap2xlsx/files/12906685/zexcel_template_test.txt)

sandraros:
For future reference, this is the Excel file:
image

and the corresponding XML part (`&L` means left, `&12` means font size 12, `&R` means right, `&P` means `&[Page]` (page number), `&N` means `&[Pages]` (number of pages), `&C` means center - For more information, see [here](https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.evenheader?view=openxml-2.8.1)):
image

sandraros (approved):
Thanks. LGTM 👍

PR #1148, opened 2023-10-25
add iv_end_row to convert_to_table

WegnerDan:
I needed a way to only get the first row of an excel sheet, so I added parameter iv_end_row to method convert_to_table of class zcl_excel_worksheet


sandraros (approved):
LGTM 👍
Thanks. It makes sense to have this new parameter, which completes the existing one (`iv_begin_row`).
NB: tested by doing a local copy of `zdemo_excel51` and adapting the call to `convert_to_table`.

PR #1183, opened 2024-02-01
Cleanup unused variables in zcl_excel_theme

sandraros:
Removing the statement ixml_element->get_attribute( 'xmlns:a' ) is anyway a good thing because getting a namespace attribute this way is not supported by IXML and it is always returning an empty string even if this "attribute" exists.


sandraros:
Obvious code fix -> approving myself.

PR #1185, opened 2024-02-06
Unused calls/variables about cl_xml_document

sandraros:
1. After calling the method `parse_string` (of `cl_xml_document`), the document instance is not used → removing the call and the unused variable.
2. The return code of method call `parse_xstring` (of `cl_xml_document`) is not used → removing the unused variable.
   NB: it's needed to keep the calls to `parse_xstring` as they are followed by the use of the instantiated document instance.


sandraros:
Obvious fix, no regression, approving myself.

PR #1186, opened 2024-02-06
Ineffective Set_Namespace_Prefix on DOM document

sandraros:
It always return an error but isn't handled → line removed


sandraros:
Method call always returns an error, seen by debug, useless line removed → no regression → approved by myself

PR #1188, opened 2024-02-14
Replace IXML with SXML (for Cloud)

sandraros:
Big PR related to #632.

- Classic abap2xlsx classes
  - Refer to ZCL/ZIF_EXCEL_XML... instead of CL/IF_IXML...
  - Use of data references with CREATE_OSTREAM_(X|C)STREAM methods

- XML
  - ZCL/ZIF_EXCEL_XML(...) - Same (...) object type names as IXML classes and interface parts used by abap2xlsx (same method and parameter names, except it's a short list of what was used by abap2xlsx till now) - Parameter types refer to ZCL/ZIF_EXCEL_XML... instead of CL/IF_IXML... but run SXML. - implementations don't contain any reference to CL/IF_IXML...
	- contains many test classes to make sure that SXML and IXML results into identical XML
	  - The tests run IXML only if ZCL_EXCEL_IXML exists (part of NOT_CLOUD package)
  - ZCL_EXCEL_XML_DOCUMENT
    - Same as CL_XML_DOCUMENT parts used by abap2xlsx

- NOT_CLOUD
  - ZCL_EXCEL_IXML - contains references to CL/IF_IXML... (it's why it's in NOT_CLOUD)
	- used by ZCL_EXCEL_XML test classes to run abap2xlsx with IXML (if IXML is installed)

Potential issues:
- Performance. Nothing optimized.
- SXML is more strict than IXML and less permissive on positioning attributes, namespace declarations, etc.
- OO design: Abuse of private instantiation/all friends.
- Probably many I didn't see...


AndreaBorgia-Abo:
On 731:
grafik

AndreaBorgia-Abo (commented):
On 731:
grafik

Additional note: with the current "main" (both library and demos), checker reports some differences in 37, 38, and 15_01. The same checks on S4H (official SAP HANA appliance) bring up no findings. Disabling the strict comparison only works within the PR branch, so no chance to check on 731.

sandraros:
Sorry, as Lars said, in the end IXML doesn't need to be replaced with SXML, it's only few classes CL_IXML to be replaced with CL_IXML_CORE, etc., all IF_IXML_* remain unchanged, so I remove my PR. It has to be recoded differently. Thanks for looking at it anyway! 👍

sandraros:
Sorry, as Lars said, in the end IXML doesn't need to be replaced with SXML, it's only few classes CL_IXML to be replaced with CL_IXML_CORE, etc., all IF_IXML_* remain unchanged, so I remove my PR. It has to be recoded differently. Thanks for looking at it anyway! 👍

PR #1190, opened 2024-02-14
CSV Initial External Date

aditya-deloitte:
Added code to set an initial external date other than the value returned by function module 'CONVERT_DATE_TO_EXTERNAL'.


sandraros (approved):
lgtm

Issue #1132, opened 2023-08-09
Cells imported as dates do not ignore time - thus the date may be rounded up to a different value than the one displayed in the spreadsheet


kjetil-kilhavn:
When a spreadsheet cell contains the date & time value 2023-08-03 14:07:11 (raw data: 45141.588321759256) with a cell format displaying only the date portion, the date value in ABAP becomes 20230804.

The time portion of the value should be ignored to ensure that the imported date value becomes the same as the value displayed in the spreadsheet.

kjetil-kilhavn:
I have a fix ready for this, but I was trying to add support for time stamps too. I think I will isolate this fix and submit a pull request after a good night's sleep.

sandraros:
Program to reproduce the bug:
  ```abap
  REPORT.
  CONSTANTS _20230803_140711 TYPE string VALUE `45141.588321759256`.

  DATA(lo_excel) = NEW zcl_excel( ).
  DATA(lo_worksheet) = lo_excel->get_active_worksheet( ).

  DATA(lo_style_date) = lo_excel->add_new_style( ).
  lo_style_date->fill->filltype = zcl_excel_style_fill=>c_fill_solid.
  lo_style_date->fill->fgcolor-rgb  = zcl_excel_style_color=>c_gray.
  lo_style_date->number_format->format_code = zcl_excel_style_number_format=>c_format_date_ddmmyyyy.

  " Due to the style, Excel displays only the date (in the user's date format,
  " i.e. according to the settings in the operating system e.g. Windows)
  lo_worksheet->set_cell( ip_columnrow = 'A1'
                          ip_value     = _20230803_140711
                          ip_style     = lo_style_date
                          ip_data_type = ' ' ).

  TYPES:
    BEGIN OF ts_read_itab,
      date TYPE d,
    END OF ts_read_itab.
  TYPES tt_read_itab TYPE STANDARD TABLE OF ts_read_itab WITH EMPTY KEY.
  DATA(gt_read_itab) = VALUE tt_read_itab( ).
  lo_worksheet->convert_to_table( EXPORTING iv_begin_row = 1
                                  IMPORTING et_data      = gt_read_itab ).
  ASSERT gt_read_itab[ 1 ]-date = '20230803'. "<== short dump because actual = '20230804'
  ```

PR #1197, opened 2024-03-24
Convert fields with date & time value correctly to just date or just time

kjetil-kilhavn:
Ignore decimals when coverting to date and ignore integer when converting to time
Fix for #1132 and also prevents runtime error when cell with time format has date & time value


sandraros:
I'm not comfortable about changing the logic in `excel_string_to_date`, in case other people have implemented a "-1 day" private fix after the call, so I added a parameter `ip_exact` to keep an ascending compatibility.

I reused the old structure of unit tests to simplify the diff. I created #1206 to reorganize the unit tests.

I created #1205 to report the time issue fixed by this PR, as the original issue mentioned only the date issue.

Note that the failed unit test is a bug (https://github.com/abaplint/transpiler/issues/1431), but it's an optional check, so the PR can be merged. EDIT: bug fixed and unit job now ok.

kjetil-kilhavn:
It's your project and therefore your decision. Feel free to merge.

sandraros (approved):

PR #1219, opened 2024-05-02
Add possibility to change the size of the comments.

### LINKED ISSUE #1221: Size of comments
### Discussed in https://github.com/abap2xlsx/abap2xlsx/discussions/1217

<div type='discussions-op-text'>

<sup>Originally posted by **RixarSAP** April 30, 2024</sup>
Hi,

Is it possible to change the size of a comment?

I've seen the code in demo program ZDEMO_EXCEL_COMMENTS.
But cannot see any way to change the size of a comment.

I'm trying to change the code of method ZCL_EXCEL_WRITER_2007-CREATE_XL_DRAWING_FOR_COMMENTS to change the size of the shape in this line: 

```
    lo_element_shape->set_attribute_ns( : name  = lc_xml_attr_id          value = lv_attr_id ),
                                          name  = lc_xml_attr_type        value = '#_x0000_t202' ),
                                          name  = lc_xml_attr_style       value = 'size:auto;width:auto;height:auto;position:absolute;margin-left:117pt;margin-top:172.5pt;z-index:1;visibility:hidden;cx=300;cy=400' ),
                                          name  = lc_xml_attr_fillcolor   value = '#ffffe1' ),
                                          name  = lc_xml_attr_oinsetmode  value = lc_xml_attr_val_auto ).
```

But doesnt work.
I've tried adding cx and cy attributes, and changing width and height... 

Any tip?

Thanks</div>

RixarSAP:
To fix #1221.

Changed class ZCL_EXCEL_COMMENT to change the method SET_TEXT adding two new OPTIONAL parameters: IP_RIGHT_COLUMN and IP_BOTTOM_ROW and add a new method: GET_POSITION.
And changed method CREATE_XL_DRAWING_FOR_COMMENTS of class ZCL_EXCEL_WRITER_2007 to read the size of the comments.

With this you can now set the size of the comments like this:

```
  DATA:
   lv_Comment              TYPE string,
   lo_comment              TYPE REF TO zcl_excel_comment.

  lv_Comment = 'Don Quijote de la Mancha.'.

  lo_comment = lo_excel->add_new_comment( ).
  lo_comment->set_text( ip_ref = 'D10' ip_text = lv_Comment ip_right_column = 5 ip_botom_row = 15 ).
  lo_worksheet->add_comment( lo_comment ).
```

image

```
  DATA:
   lv_Comment              TYPE string,
   lo_comment              TYPE REF TO zcl_excel_comment.

  lv_Comment = 'En un lugar de la Mancha, de cuyo nombre no quiero acordarme, no ha mucho tiempo que vivía un hidalgo de los de lanza en astillero, adarga antigua, rocín flaco y galgo corredor.'.

  lo_comment = lo_excel->add_new_comment( ).
  lo_comment->set_text( ip_ref = 'D10' ip_text = lv_Comment  ip_right_column = 10 ip_bottom_row = 30 ).
  lo_worksheet->add_comment( lo_comment ).
```

image

image

### DIFF
--- a/src/zcl_excel_comment.clas.abap
+++ b/src/zcl_excel_comment.clas.abap
@@ -5,41 +5,43 @@ CLASS zcl_excel_comment DEFINITION
 
   PUBLIC SECTION.
 
-    CONSTANTS default_right_column TYPE i VALUE 4.          "#EC NOTEXT
-    CONSTANTS default_bottom_row TYPE i VALUE 15.           "#EC NOTEXT
+    CONSTANTS default_right_column TYPE i VALUE 4.  "#EC NOTEXT
+    CONSTANTS default_bottom_row   TYPE i VALUE 15. "#EC NOTEXT
 
     METHODS constructor .
-    METHODS get_name
+    METHODS get_bottom_row
       RETURNING
-        VALUE(r_name) TYPE string .
+        VALUE(rp_result) TYPE i .
     METHODS get_index
       RETURNING
         VALUE(rp_index) TYPE string .
+    METHODS get_name
+      RETURNING
+        VALUE(r_name) TYPE string .
     METHODS get_ref
       RETURNING
         VALUE(rp_ref) TYPE string .
+    METHODS get_right_column
+      RETURNING
+        VALUE(rp_result) TYPE i .
     METHODS get_text
       RETURNING
         VALUE(rp_text) TYPE string .
     METHODS set_text
       IMPORTING
-        !ip_text TYPE string
-        !ip_ref  TYPE string OPTIONAL
+        !ip_text         TYPE string
+        !ip_ref          TYPE string OPTIONAL
         !ip_right_column TYPE i OPTIONAL
-        !ip_bottom_row TYPE i OPTIONAL .
-    METHODS get_position
-      EXPORTING
-        !ep_right_column TYPE i
-        !ep_bottom_row TYPE i .
+        !ip_bottom_row   TYPE i OPTIONAL .
 
   PROTECTED SECTION.
   PRIVATE SECTION.
 
+    DATA bottom_row TYPE i .
     DATA index TYPE string .
     DATA ref TYPE string .
-    DATA text TYPE string .
     DATA right_column TYPE i .
-    DATA bottom_row TYPE i .
+    DATA text TYPE string .
 ENDCLASS.
 
 
@@ -49,6 +51,11 @@ CLASS zcl_excel_comment IMPLEMENTATION.
 
   METHOD constructor.
 
+  ENDMETHOD.
+
+
+  METHOD get_bottom_row.
+    rp_result = bottom_row.
   ENDMETHOD.
 
 
@@ -67,6 +74,11 @@ CLASS zcl_excel_comment IMPLEMENTATION.
   ENDMETHOD.
 
 
+  METHOD get_right_column.
+    rp_result = right_column.
+  ENDMETHOD.
+
+
   METHOD get_text.
     rp_text = me->text.
   ENDMETHOD.
@@ -79,30 +91,17 @@ CLASS zcl_excel_comment IMPLEMENTATION.
       me->ref = ip_ref.
     ENDIF.
 
-  IF ip_right_column IS SUPPLIED.
-    me->right_column = ip_right_column.
-  ENDIF.
+    IF ip_right_column IS NOT INITIAL.
+      me->right_column = ip_right_column.
+    ELSE.
+      me->right_column = default_right_column.
+    ENDIF.
 
-  IF ip_bottom_row IS SUPPLIED.
-    me->bottom_row = ip_bottom_row.
-  ENDIF.
-
+    IF ip_bottom_row IS NOT INITIAL.
+      me->bottom_row = ip_bottom_row.
+    ELSE.
+      me->bottom_row = default_bottom_row.
+    ENDIF.
   ENDMETHOD.
-
-METHOD get_position.
-
-  IF right_column IS NOT INITIAL.
-    ep_right_column = right_column.
-  ELSE.
-    ep_right_column = default_right_column. "Default right_column
-  ENDIF.
-
-  IF bottom_row IS NOT INITIAL.
-    ep_bottom_row = bottom_row.
-  ELSE.
-    ep_bottom_row = default_bottom_row. "Default bottom_row
-  ENDIF.
-
-ENDMETHOD.
 
 ENDCLASS.
--- a/src/zcl_excel_comment.clas.abap
+++ b/src/zcl_excel_comment.clas.abap
@@ -5,6 +5,9 @@ CLASS zcl_excel_comment DEFINITION
 
   PUBLIC SECTION.
 
+  constants DEFAULT_WIDTH type I value 2. "#EC NOTEXT
+  constants DEFAULT_HEIGHT type I value 15. "#EC NOTEXT
+
     METHODS constructor .
     METHODS get_name
       RETURNING
@@ -22,12 +25,23 @@ CLASS zcl_excel_comment DEFINITION
       IMPORTING
         !ip_text TYPE string
         !ip_ref  TYPE string OPTIONAL .
+  methods SET_SIZE
+    importing
+      !IP_WIDTH type I optional
+      !IP_HEIGHT type I optional .
+  methods GET_SIZE
+    exporting
+      !EP_WIDTH type I
+      !EP_HEIGHT type I .
+
   PROTECTED SECTION.
   PRIVATE SECTION.
 
     DATA index TYPE string .
     DATA ref TYPE string .
     DATA text TYPE string .
+  data WIDTH type I .
+  data HEIGHT type I .
 ENDCLASS.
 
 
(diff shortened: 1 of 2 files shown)


sandraros:
Thank you. If you don't mind, could you adjust the pretty print e.g. "[Commit suggestion](https://github.com/abap2xlsx/abap2xlsx/pull/1219/files)" repeatedly till it's okay?
image

RixarSAP:
> Thank you. If you don't mind, could you adjust the pretty print e.g. "[Commit suggestion](https://github.com/abap2xlsx/abap2xlsx/pull/1219/files)" repeatedly till it's okay? image

I think is now ok. Let me know if I need to fix anything else.

sandraros:
Nice! I'll review, test and merge it as quickly as I can, I guess today or tomorrow.

RixarSAP:
> Nice! I'll review, test and merge it as quickly as I can, I guess today or tomorrow.

Great ! A pleasure to be able to contribute.
I have some other ideas to implement, I will create the corresponding forks...

darnoc312:
Very good idea. But i would prefer two new optional parameters for method set_text instead of a new method set_size. According to parameter ref, which is set by set_text but gets its own get method. Of course this is only my opinion. And the default width in the old string seems to be 4 and not 2.

RixarSAP:
> Very good idea. But i would prefer two new optional parameters for method set_text instead of a new method set_size. According to parameter ref, which is set by set_text but gets its own get method. Of course this is only my opinion. And the default width in the old string seems to be 4 and not 2. 

@darnoc312 
I was thinking about doing it like this at first, all in the SET_TEXT method.
The problem is that I need to recover the size later. and The method  GET_TEXT only has one returning type parameter. I didn't like changing its interface. So I needed a new GET_SIZE method anyway, and if I have a GET it would seem okay to also have a SET.

I can change it if you see it better that way.

About the default width you're right, I'm going to change it from 2 o 4.
Thanks for comment.

EDIT: I'm thinking is better as you said. Changed the file.

RixarSAP:
> > Very good idea. But i would prefer two new optional parameters for method set_text instead of a new method set_size. According to parameter ref, which is set by set_text but gets its own get method. Of course this is only my opinion. And the default width in the old string seems to be 4 and not 2.
> 
> @darnoc312 I was thinking about doing it like this at first, all in the SET_TEXT method. The problem is that I need to recover the size later. and The method GET_TEXT only has one returning type parameter. I didn't like changing its interface. So I needed a new GET_SIZE method anyway, and if I have a GET it would seem okay to also have a SET.
> 
> I can change it if you see it better that way.
> 
> About the default width you're right, I'm going to change it from 2 o 4. Thanks for comment.
> 
> EDIT: I'm thinking is better as you said. Changed the file to modify method SET_TEXT.

@darnoc312 I've reset the default width to 2 again...

I've changed other numbers in the anchor string and I think is better the width of 2.

image

sandraros:
The new code should maintain the ascending compatibility, which is not the case as per your screenshot. Could you check why the "new default 4 15" is not the same as the "old default 4 15"?

RixarSAP:
> The new code should maintain the ascending compatibility, which is not the case as per your screenshot. Could you check why the "new default 4 15" is not the same as the "old default 4 15"?

Ok, I think now its solved.

image

sandraros:
There's a little misunderstanding or at least a misnaming of the anchor parameters. It's not "width" and "height", it's "right column" and "bottom row".

The eight numbers are four pairs which indicate the positions of:
- left column
- top row
- right column
- bottom row
The first number of each pair is relative to the cell where the comment is attached.
The second number of each pair is a complementary shift expressed as a number of pixels.

Reference: https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.vml.spreadsheet.anchor?view=openxml-3.0.1

We might choose better names for now, instead of `width`, `height`, `get_size` (e.g. `get_position` would be probably better).

Also, `get_size` doesn't make sense in abap2xlsx currently, it's only needed by your custom logic, so is it possible to give it a reason to exist in abap2xlsx? For instance, is it possible to implement it in `zcl_excel_reader_2007`? It's possibly complex, so it's just a question.

RixarSAP:
> There's a little misunderstanding or at least a misnaming of the anchor parameters. It's not "width" and "height", it's "right column" and "bottom row".
> 
> The eight numbers are four pairs which indicate the positions of:
> 
> * left column
> * top row
> * right column
> * bottom row
>   The first number of each pair is relative to the cell where the comment is attached.
>   The second number of each pair is a complementary shift expressed as a number of pixels.
> 
> Reference: https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.vml.spreadsheet.anchor?view=openxml-3.0.1
> 
> We might choose better names for now, instead of `width`, `height`, `get_size` (e.g. `get_position` would be probably better).
> 
> Also, `get_size` doesn't make sense in abap2xlsx currently, it's only needed by your custom logic, so is it possible to give it a reason to exist in abap2xlsx? For instance, is it possible to implement it in `zcl_excel_reader_2007`? It's possibly complex, so it's just a question.

Ohh, thanks for the reference, I was looking for it this morning but couldnt find it.
I was just playing and testing with the values.

So basically we have this:

image

The offsets are in pixels, so, It depends on the screen resolution?
I think we can leave all these offset set to 0. 
I've made a quick test with the following anchor ( '2, 0, 11, 0, 15, 0, 30, 0' ) and works fine and create a big comment.
So I think we need 4 parameters, just like when we call a screen with STARTING AT, ENDING AT...

I'll check the code tomorrow and make some changes (now its about 11 P.M. here...)

About the method GET_SIZE well... I need a method in class zcl_excel_comments to recover these values, because in method CREATE_XL_DRAWING_FOR_COMMENTS of class ZCL_EXCEL_WRITER_2007 is where the shape of the comment is drawn, there is a loop to all the comments created in the worksheet:

```
  lo_iterator = lo_comments->get_iterator( ).
  WHILE lo_iterator->has_next( ) EQ abap_true.
    lv_index = sy-index.
    lo_comment ?= lo_iterator->get_next( ).
```

So I need this method to get these values. 
I'll rethink all tomorrow and change all the names of the attributes and methods.

Thanks for the help !

RixarSAP:
> There's a little misunderstanding or at least a misnaming of the anchor parameters. It's not "width" and "height", it's "right column" and "bottom row".
> 
> The eight numbers are four pairs which indicate the positions of:
> 
> * left column
> * top row
> * right column
> * bottom row
>   The first number of each pair is relative to the cell where the comment is attached.
>   The second number of each pair is a complementary shift expressed as a number of pixels.
> 
> Reference: https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.vml.spreadsheet.anchor?view=openxml-3.0.1
> 
> We might choose better names for now, instead of `width`, `height`, `get_size` (e.g. `get_position` would be probably better).
> 
> Also, `get_size` doesn't make sense in abap2xlsx currently, it's only needed by your custom logic, so is it possible to give it a reason to exist in abap2xlsx? For instance, is it possible to implement it in `zcl_excel_reader_2007`? It's possibly complex, so it's just a question.

@sandraros In the end I made the changes as you said.
The goal of this feature is to have the possibility to change the size of the comment, not to move the comment arround the sheet...  so I've left all the numbers like in the original code (to have backward compatibility) and play only with these two values: right_column and botton_row.

Hope its ok now.

sandraros:
In the end, I don't like `get_position` either ;-) I propose these little changes (see the latest commit):
- One method per attribute: `get_bottom_row` and `get_right_column`.
- Set the value of the attributes during `set_text`, not in the `get` method(s).
- I added a `condense` to remove one space, just to make the result of `zdemo_excel_comments` exactly the same as before.
- Little pretty print and change of the order of attributes/methods (I find the alphabetical order a quick way to find the names when scrolling / I'm trying to adhere to the ABAP Clean Code principles).

What do you think?

> About the method GET_SIZE well... I need a method in class zcl_excel_comments to recover these values...

Oops sorry, yes of course.

RixarSAP:
> In the end, I don't like `get_position` either ;-) I propose these little changes (see the latest commit):
> 
> * One method per attribute: `get_bottom_row` and `get_right_column`.
> * Set the value of the attributes during `set_text`, not in the `get` method(s).
> * I added a `condense` to remove one space, just to adhere to make the result of `zdemo_excel_comments` exactly the same as before.
> * Little pretty print and change of the order of attributes/methods (I find the alphabetical order a quick way to find the names when scrolling / I'm trying to adhere to the ABAP Clean Code principles).
> 
> What do you think?
> 
> > About the method GET_SIZE well... I need a method in class zcl_excel_comments to recover these values...
> 
> Oops sorry, yes of course.

Oh looks great to me. I've seen the code from my mobile but look all fine and make sense this way.
I'll replicate the changes manually in my system this monday (I cannot use the report zgithub_stsndalone in my system due to my old abap versión )

Thanks for your time !

RixarSAP:
Replicated last changes and works all great !
Its now all ok?

sandraros:
Merged, thank you very much!

RixarSAP:
> Merged, thank you very much!

oh, great ! thanks for your support !

I have some other ideas, i'll create a discussion before to create a new pull request to see if like them.

Issue #1081, opened 2023-01-28
Read of file generated by abap2xlsx fails "Please provide either row and column, or cell reference" 


sandraros:
Steps to reproduce:
1. Create this program:
    ```abap
    REPORT.
    DATA: gc_save_file_name TYPE string VALUE 'test.xlsx'.
    INCLUDE zdemo_excel_outputopt_incl.
    START-OF-SELECTION.
      DATA(lo_excel) = NEW zcl_excel( ).
      DATA(lo_worksheet) = lo_excel->get_active_worksheet( ).
      DATA(style_large_bold) = lo_excel->add_new_style( ).
      lo_excel->set_default_style( style_large_bold->get_guid( ) ).
      lo_worksheet->set_cell( ip_columnrow = 'A2' ip_value = 'test' ).
      DATA(lo_writer) = CAST zif_excel_writer( NEW zcl_excel_writer_2007( ) ).
      DATA(xstring) = lo_writer->write_file( lo_excel ).
      DATA(lo_reader) = NEW zcl_excel_reader_2007( ).
      DATA(lo_excel_2) = lo_reader->zif_excel_reader~load( i_excel2007 = xstring ). " <=== Exception
      lo_excel = lo_excel_2.
      lcl_output=>output( cl_excel = lo_excel ).
    ```
1. Run the program > Choose "direct display":
   <img src="https://user-images.githubusercontent.com/34005250/215270470-e0e3dfee-443c-4a54-8583-504deca97695.png" width="250">
1. There is a short dump while calling the method `LOAD` due to uncaught exception `zcx_excel` with `error` text "Please provide either row and column, or cell reference"
   <img src="https://github.com/abap2xlsx/abap2xlsx/assets/34005250/78b77460-225b-4df7-a60b-bd535985272e" width="500">
   <img src="https://github.com/abap2xlsx/abap2xlsx/assets/34005250/08115c4d-baba-44a4-85ad-1c2617c05c7f" width="500">
   <img src="https://github.com/abap2xlsx/abap2xlsx/assets/34005250/b92645c2-a6c1-43bb-9d7d-af37d83c9abb" width="500">
   <img src="https://github.com/abap2xlsx/abap2xlsx/assets/34005250/d37edaaf-1d0a-4ffa-b49e-3b8f18a9ba14" width="500">
   Link to the code:
[...]

PR #1222, opened 2024-05-03
Dummy lines in WRITER's method create_xl_sheet should not create 'C' node without cell coords

darnoc312:
Hallo,

solves #1081.

You know: The aim of using dummy lines in this code is to avoid omitting any row class data.

There are two kinds of dummy lines.
The start and end lines containing the dummy cell content and the lines created in the WHILE Loop.
Solution:
The start and end lines must be handled more exactly like the other dummy lines.

That means:
1) Their row data class must be checked for relevant values, too. That's why the existing CHECK now gets its own surrounding IF.
2) And they get no 'C' node in the 'ROW' node, too That's why i included a new CHECK before creating this node.
   For explanation: The other dummy lines create ROW data in the WHILE loop an cannot get to this point.

Note: It is intended to use ls_sheet_content in the new CHECK line. If continuing the loop at this point the field symbol <ls_sheet_content> may be not or wrong assigned.

Furthermore i would prefer renaming <ls_sheet_content> to ls_sheet_content from this point down to the ENDLOOP.
The usage of the field-symbol is not caused by any relation to the WHILE loop but because the LOOP starts in the early days of ABAP2XLSX with ASSIGNING TO <ls_sheet_content>. But now INTO ls_sheet_content is used.
To avoid confusion this is not included yet.

### DIFF
--- a/src/zcl_excel_writer_2007.clas.abap
+++ b/src/zcl_excel_writer_2007.clas.abap
@@ -4343,9 +4343,9 @@ CLASS zcl_excel_writer_2007 IMPLEMENTATION.
            ls_sheet_content-cell_value = lc_dummy_cell_content.
 *Check if empty row is really necessary - this is basically the case when we have information in row_dimension
           lo_row_empty = io_worksheet->get_row( lv_current_row ).
-          CHECK lo_row_empty->get_row_height( )                 >= 0         OR
-                lo_row_empty->get_collapsed( io_worksheet )      = abap_true OR
-                lo_row_empty->get_outline_level( io_worksheet )  > 0         OR
+          CHECK lo_row_empty->get_row_height( )                 >= 0          OR
+                lo_row_empty->get_collapsed( io_worksheet )      = abap_true  OR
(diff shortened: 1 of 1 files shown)


sandraros:
I propose one PR to fix the bug. Isn't it just one line to be added? And then use another PR to make the code more clear, but it would require to refactor it a lot. Currently, I'm sorry, but I don't see the added value of your PR, the code is as obscure as previously. A code more clear should not rely on dummy cells.

darnoc312:
I didn't know you would agree to a big change.
But of course that is feasible.
New PR follows. I am closing this one.

sandraros:
I don't really agree to a big internal change which would require a lot of time to think and test. If you want to do it, I prefer to have several small PRs, each one being easy to understand and no risk "by design". Note that the overall code of abap2xlsx is quite messy, with messy design which cannot be improved (or abap2xlsx will lose its ascending compatibility), so what is the advantage to improve this part?

sandraros:
I just proposed another solution -> #1225

PR #1224, opened 2024-05-09
Added field WIDTH in field catalog of method BIND_TABLE

RixarSAP:
I've added a new field WIDTH in structure ZEXCEL_S_FIELDCATALOG.
Filling this field you can set a WIDTH to this column using method BIND_TABLE.

To test it you can modify the report ZDEMO_EXCEL10, setting a value to this field.

The goal of this change is to not call the method SET_COLUMN_WIDTH repeatedly after the call of BIND_TABLE. 
You can set a width in the fieldcat like in the ABAP ALVs.


sandraros (approved):
Looks good to me.

darnoc312:
Repeatedly ? Not more or less than before: 1 time per column.
But nevertheless: In my opinion this change was not as well as it seems.
The column width is an attribute of the worksheet and not of any table in the worksheet.
For example: If you create two tables one below the other in one worksheet,
there will be at least one worksheet column that belongs to both tables.
Which table column width should win ? 
In the previous code of ZDEMO_EXCEL10 the width 13 was assigned to column 'B' of the worksheet.
Then this is clearly intended even if there would be more than one table.

PR #1225, opened 2024-05-11
Support of empty cell address by LOAD (workbook)

### LINKED ISSUE #1081: Read of file generated by abap2xlsx fails "Please provide either row and column, or cell reference"

sandraros:
Fix #1081


sandraros:
@AndreaBorgia-Abo Could you review this fix please? There was several possible fixes, one via the writer and one via the reader. I feel that fixing the reader is a much better way. Thank you.

AndreaBorgia-Abo (approved):
LGTM!

Issue #1235, opened 2024-06-08
Worksheet's method set_title doesn't handle escaped titles and sheetlocal ranges 


darnoc312:
Hello,

please try the following demo program similar to ZDEMO_EXCEL8 which sets title at the end again

```abap
REPORT zdemo_excel8_test.

DATA: lo_excel                TYPE REF TO zcl_excel,
      lo_worksheet            TYPE REF TO zcl_excel_worksheet,
      lo_range                TYPE REF TO zcl_excel_range.

DATA: lv_title          TYPE zexcel_sheet_title.

CONSTANTS: gc_save_file_name TYPE string VALUE '08_Range.xlsx'.
INCLUDE zdemo_excel_outputopt_incl.


START-OF-SELECTION.

  CREATE OBJECT lo_excel.

  " Get active sheet
  lo_worksheet = lo_excel->get_active_worksheet( ).
  lv_title = 'Escaped title text'.                                        "Title for testing
  lo_worksheet->set_title( lv_title ).
  lo_range = lo_excel->add_new_range( ).
  lo_range->name = 'range'.
  lo_range->set_value( ip_sheet_name    = lv_title
                       ip_start_column  = 'C'
                       ip_start_row     = 4
                       ip_stop_column   = 'C'
                       ip_stop_row      = 8 ).


  lo_worksheet->set_cell( ip_row = 4 ip_column = 'C' ip_value = 'Apple' ).
  lo_worksheet->set_cell( ip_row = 5 ip_column = 'C' ip_value = 'Banana' ).
  lo_worksheet->set_cell( ip_row = 6 ip_column = 'C' ip_value = 'Blueberry' ).
  lo_worksheet->set_cell( ip_row = 7 ip_column = 'C' ip_value = 'Ananas' ).
  lo_worksheet->set_cell( ip_row = 8 ip_column = 'C' ip_value = 'Grapes' ).

  " Define another Range with a name longer than 40 characters that
  " tests the fix of issue #168 ranges namescan be only up to 20 chars

  lo_range = lo_excel->add_new_range( ).
  lo_range->name = 'A_range_with_a_name_that_is_longer_than_20_characters'.
  lo_range->set_value( ip_sheet_name    = lv_title
                       ip_start_column  = 'D'
                       ip_start_row     = 4
                       ip_stop_column   = 'D'
                       ip_stop_row      = 5 ).
[...]

PR #1236, opened 2024-06-08
Worksheet's method set_title doesn't handle escaped titles and sheetlocal ranges

darnoc312:
solves #1235 

And of course the range named 'A_range_with_sheet_visibility' needs both fixes.


sandraros (approved):
Perfect, thank you!

PR #1239, opened 2024-06-14
Yet another autofilter problem using converter and ALV

### LINKED ISSUE #1238: Yet another autofilter problem using converter and ALV
Use ZDEMO_EXCEL32 and select one column filter with one value to avoid side effect with the other autofilter issue.
In my example i chosed '00000111' for AGENCYNUM. You get the following result:

[32_Export_Convert #1.xlsx](https://github.com/user-attachments/files/15834749/32_Export_Convert.1.xlsx)

XML Autofilter's ref in xl/worksheets/sheet1.xml is wrongly T1:T1 and nothing is hidden.
Converter's method set_autofilter_area sets row_start to 1. row_end remains 0.
Currently autofilter's method validate_area then exchanges these values wrongly. row_start = 0, row_end = 1.
Then the following validation code sets row_start because of 0 to 1. Final result: row_start = 1 and row_end = 1.

The bug is the change of autofilters method validate_area with PR #822. The change there is useful but why the changed code has been moved ?
It seems to be obvious, that the behaviour of row and column should be similar in this method. But since this PR that has no longer been the case.

So the first fix is moving this code to its previous place.
Then you get the following result:

[32_Export_Convert #2.xlsx](https://github.com/user-attachments/files/15834752/32_Export_Convert.2.xlsx)

XML Autofilter's ref is T1:T52 and the header is wrongly hidden.
Converter's method set_autofilter_area sets row_start to 1. row_end remains 0.
Then autofilter's method validate_area sets row_end because of 0 to highest_row correctly and no further exchange will happen.
Final result: row_start = 1 and row_end = 52.

The next bug is the setting of row_start in converter's method set_autofilter_area to 1 instead to the value of converter's field w_row_int (The converter call in ZDEMO_EXCEL32 starts at row 2 and column 2).
After this is also fixed you get the following result:

[32_Export_Convert #3.xlsx](https://github.com/user-attachments/files/15834756/32_Export_Convert.3.xlsx)

XML Autofilter's ref is now correctly T2:T52.
Converter's method set_autofilter_area sets row_start to 2 (the value of w_row_int). row_end remains 0.
Then autofilter's method validate_area sets row_end because of 0 to highest_row correctly and no further exchange will happen.
Final result: row_start = 2 and row_end = 52.

darnoc312:
fix #1238

The change of autofilter's method validate_area is the minimum change.
In my opinion it should even look like this:

```abap
    IF filter_area-row_start < 1 OR
       filter_area-row_start > l_row.                            "new line
      filter_area-row_start = 1.
    ENDIF.
    IF filter_area-col_start < 1 OR
       filter_area-col_start > l_col.                            "new line
      filter_area-col_start = 1.
    ENDIF.
    IF filter_area-row_end > l_row OR
       filter_area-row_end < 1.
      filter_area-row_end = l_row.
    ENDIF.
    IF filter_area-col_end > l_col OR
       filter_area-col_end < 1.
      filter_area-col_end = l_col.
    ENDIF.

    IF filter_area-row_start > filter_area-row_end.
      ls_original_filter_area = filter_area.
      filter_area-row_start = ls_original_filter_area-row_end.
      filter_area-row_end   = ls_original_filter_area-row_start.
    ENDIF.
    IF filter_area-col_start > filter_area-col_end.              "now more similar to row
      ls_original_filter_area = filter_area.
      filter_area-col_start = ls_original_filter_area-col_end.
      filter_area-col_end   = ls_original_filter_area-col_start.
    ENDIF.                                                       "end of similar to row
```


bigld (approved):
LGTM

sandraros:
I just tested and nothing has changed, it's still the old result without the filter. Could you verify please?

darnoc312:
Hello,

could you provide your result, please ?
With the previous behaviour, autofilter may be wrong, but at least there should be a visible autofilter icon, may be at the wrong line.

@Bigld approved this and i think he made a successfull test.

I have attached my last result here:
[Uploading 32_Export_Convert.xlsx…]()
Thank you.

sandraros:
Same actual result as https://github.com/abap2xlsx/abap2xlsx/issues/1238#issuecomment-2169278154, I couldn't get the expected result.
Just tested now with the latest commits of abap2xslx and demos:

image

darnoc312:
Hello,

please refer to my old PR #1131.
These are the important lines:
If you use for example ZDEMO_EXCEL32 for testing be careful using ZDEMO_EXCEL33 at the same time.
Unfortunately ZDEMO_EXCEL33 calls method zcl_excel_converter->set_option instead of using the is_option parameter of the immediately following zcl_excel_converter->convert method call.
So these options, which don't contain filter = 'X', remain and the defaults of method zcl_excel_converter->init_option are not used for any further convert method call with ALV (like in ZDEMO_EXCEL32) until the 'EXCEL_username' record is deleted from database INDX.

If you ever have used ZDEMO_EXCEL33 before, you have no filter in ZDEMO_EXCEL32 until you delete this record in INDX.
I changed my ZDEMO_EXCEL33 a long time ago to avoid this problem. And now I didn't immediately think of it.

sandraros:
Could you correct #1238 to explain how to reproduce the issue please?

EDIT: okay, got it. I fixed ZDEMO_EXCEL32, now I can see that the issue is fixed.

sandraros (approved):
LGTM

PR #1255, opened 2024-08-10
Fix bug wrong format for conditional formatting of text functions (begins with, etc.)

### LINKED ISSUE #1254: Wrong format (color, etc.) for conditional formatting of text functions (begins with, etc.)
Hello,

Please change lv_style_1_guid in line 218 of ZDEMO_EXCEL27 to lv_style_2_guid for testing purpose:

```abap
    ls_textfunction-cell_style       = lv_style_2_guid.  "lv_style_1_guid
```
The result should be red color for the texfunctions (Row 6 of excel file) but it is green as before.

darnoc312:
fix #1254

The wrong field is used in method add_conditional_formatting in local class lcl_create_xl_sheet (currently line 1174).
Instead of cell style (GUID) from texfunction's data, cell style from last cellis' data is used. That is green.


sandraros:
The proposed modification isn't sufficient to fix the minimal reproducible program I have added in #1254. I have added a commit to your PR. Could you check it please? Thank you.

darnoc312:
Yes of course that was missing. In ZDEMO_EXCEL27 lv_style_2_guid was already used by the cellis conditon for 'Carp'.
So this guid already exists in table styles_cond_mapping and it works in this case without the new call of method create_dxf_style.

But this gives the occasion to solve another problem in this context:

- ZDEMO_EXCEL27 creates 14 conditional styles (all of them dxfid concerning types cellis or textfunction).
Because of your change the method create_dxf_style now is called 14 times, too (and not less as before).
But these 14 conditional styles share only 2 different (normal) style guids.
This method maintains the table styles_cond_mapping.
Currently lv_style_1_guid is inserted 9 times (9 x cellis) and lv_style_2_guid 5 times (1 x cellis and 4 x textfunction).
Each of them with the same content for style and dxf.
lcl_create_xl_sheet's method add_conditional_formatting reads this table by guid. So only one entry per guid is enough.
I inserted a CHECK at the beginning of this method to avoid these duplicates.

- I cleaned up the weird use of XML ID fields (starting from 0) as index fields (starting from 1).
When Alessandro Iannacci inserted the font part he did not continue this way and implemented the field lv_index.
Now i am using this field to read the tables it_cellxfs and it_fills, too.
Furthermore the field style in table styles_cond_mapping is filled really with the corresponding value of table styles_mapping.
Of course now this table is read by this value, too.
Debugging is a little confusing if the style value in table styles_cond_mapping for a certain guid is not equal to its value in table styles_mapping.

- The last APPEND_CHILD at the end of this method is now called only if lo_sub_element really exists.

sandraros (approved):
LGTM thank you, merging! 👍

PR #1262, opened 2024-08-17
Fix default cell format to not use UTCLONG

### LINKED ISSUE #1261: Cell with formula has format like UTCLONG (j/m/aa h:mm) instead of standard
A simple formula defined in a cell is assigned the time format "j/m/aa h:mm" corresponding to UTCLONG, instead of the Excel standard format.

This happens since abap2xlsx supports the new ABAP type UTCLONG (#1253), if the program runs in a system before ABAP 7.54 (where UTCLONG does not exist yet).

Code to reproduce:
```abap
REPORT z_issue_1261.
CONSTANTS gc_save_file_name TYPE string VALUE 'Issue_1261.xlsx'.
START-OF-SELECTION.
  lo_worksheet->set_cell( ip_columnrow = 'A1'
                          ip_formula   = 'IF(1=1,2,3)' ).
```

Actual result:

image

Expected result:

image

sandraros:
Fix #1261

### DIFF
--- a/src/zcl_excel_worksheet.clas.abap
+++ b/src/zcl_excel_worksheet.clas.abap
@@ -709,7 +709,8 @@ CLASS zcl_excel_worksheet DEFINITION
 *"* do not include other source files here!!!
     TYPES ty_table_settings TYPE STANDARD TABLE OF zexcel_s_table_settings WITH DEFAULT KEY.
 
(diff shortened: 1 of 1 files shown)


AndreaBorgia-Abo (approved):
before)
NPL 7.52: wrong result
A4H 7.5H: expected result
after)
NPL: expected result
A4H: expected result

LGTM

PR #1268, opened 2024-09-02
Options to skip hidden rows and columns for CSV format

oliver-huetkoeper:
Add options to skip hidden rows and columns when saving the data in CSV format. This is especially useful, when converting an ALV to an Excel object and saving it as CSV file. ALV layouts often display only selected columns.

### DIFF
--- a/src/zcl_excel_writer_csv.clas.abap
+++ b/src/zcl_excel_writer_csv.clas.abap
@@ -125,6 +125,7 @@ CLASS ZCL_EXCEL_WRITER_CSV IMPLEMENTATION.
     SELECT ddtext INTO ls_format-attvalue FROM dd07t WHERE domname    = 'XUDATFM'
                                                        AND ( ddlanguage = sy-langu OR ddlanguage = 'E' ).
       ls_format-cmpname = 'DATE'.
+      ls_format-attvalue = ls_format-attvalue(10). " Ignore description, only use pattern
       CONDENSE ls_format-attvalue.
       CONCATENATE '''' ls_format-attvalue '''' INTO ls_format-attvalue.
       APPEND ls_format TO lt_format.
--- a/src/zcl_excel_writer_csv.clas.abap
+++ b/src/zcl_excel_writer_csv.clas.abap
@@ -165,6 +165,9 @@ CLASS ZCL_EXCEL_WRITER_CSV IMPLEMENTATION.
     IF skip_hidden_rows = abap_true.
 * --- Retrieve autofilters (to identify hidden rows)
       lo_autofilter = excel->get_autofilters_reference( )->get( io_worksheet = lo_worksheet ).
+      IF lo_autofilter IS NOT INITIAL.
+        lo_autofilter->get_filter_area( ). " trigger filter area validation
+      ENDIF.
     ENDIF.
 
     lv_row = 1.
--- a/src/zcl_excel_writer_csv.clas.abap
+++ b/src/zcl_excel_writer_csv.clas.abap
@@ -31,10 +31,10 @@ CLASS zcl_excel_writer_csv DEFINITION
       IMPORTING
         !ip_value TYPE char10 DEFAULT c_default .
     CLASS-METHODS set_skip_hidden_rows
-      IMPORTING 
+      IMPORTING
         !ip_value TYPE abap_bool.
     CLASS-METHODS set_skip_hidden_columns
-      IMPORTING 
+      IMPORTING
         !ip_value TYPE abap_bool.
 *"* protected components of class ZCL_EXCEL_WRITER_CSV
 *"* do not include other source files here!!!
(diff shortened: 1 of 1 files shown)


sandraros:
Thank you. I'll review it this week.

sandraros:
I have created the issue #1274 with a program for testing the enhancement.

sandraros:
I ran the demo program indicated in the issue + the new lines below, only the rows of USA companies should be visible (AA and CO) but they are not hidden. Could you check what's going on? Thank you.

image

```abap
  DATA(lo_excel_writer) = cast zif_excel_writer( new zcl_excel_writer_csv( ) ).
  zcl_excel_writer_csv=>set_delimiter( p_sep ).
  zcl_excel_writer_csv=>set_enclosure( '''' ).
  zcl_excel_writer_csv=>set_endofline( cl_abap_char_utilities=>cr_lf ).
  zcl_excel_writer_csv=>set_initial_ext_date( '' ).
  zcl_excel_writer_csv=>set_skip_hidden_rows( abap_true ).     " <=== new line to test the pull request
  zcl_excel_writer_csv=>set_skip_hidden_columns( abap_true ).  " <=== new line to test the pull request
  DATA(lv_file) = lo_excel_writer->write_file( lo_excel ).
```

oliver-huetkoeper:
I ran the test report in my system and it works fine. The CSV file only contains two data rows (+header).

In general my system is not on the latest abap2xlsx version and I cannot update it. But the CSV writer class is the same. I also checked some other relevant classes (zcl_excel_converter_salv_table, …) and found no relevant differences.

Can you please set a break-point in method create_csv of the csv writer and check if lo_autofilter->mt_filters contains the filter (column=4, t_values contains „USD“)?

darnoc312:
Autofilter needs a validation to work.
But the method autofilter->validate_area is private and is called by these methods only
image
None of them is used here. Please add the following lines after retrieving the autofilter object:
```abap
  lo_autofilter = excel->get_autofilters_reference( )->get( io_worksheet = lo_worksheet ).
  IF lo_autofilter IS BOUND.
    lo_autofilter->get_filter_area( ).  "call filter area validation
  ENDIF.
```
Not to get the filter area but simply to call the validation.

sandraros:
> Can you please set a break-point in method create_csv of the csv writer and check if lo_autofilter->mt_filters contains the filter (column=4, t_values contains „USD“)?

Yes, it does contain it.

The issue is with `filter_area-row_end` in method `is_row_hidden` of `zcl_excel_autofilter`, it contains zero, which means that the filter applies to a "range up to the row 0" i.e. it doesn't apply to anything.

The extra line proposed by @darnoc312 would solve the issue.

I guess we could go for this solution.

NB: hopefully in the future we can get rid of this obscure "validation" in the class `zcl_excel_autofilter`.

darnoc312:
Please rethink about your algorithm.
For performance reasons method is_row_hidden should be called one time per row only.
But currently this method is called for every cell.
Your CONTINUE doesn't skip a row. It skips to the next cell only.

sandraros:
Concerning the legacy content of the method `create_csv` of `zcl_excel_writer_csv`, it would be best to refactor its logic to have two loops, one loop for the rows, one loop for the cells of each row. The original internal table `sheet_content` is sorted by row then by column, so it could be used directly. I don't understand why the cells with formulas should not be output in the CSV... (`DELETE lt_cell_data WHERE cell_formula IS NOT INITIAL.`)

oliver-huetkoeper:
> Please rethink about your algorithm. For performance reasons method is_row_hidden should be called one time per row only. But currently this method is called for every cell. Your CONTINUE doesn't skip a row. It skips to the next cell only.

I though about deleting the rows/columns from the internal table. But there is some „add empty rows/columns“ logic which would break. I wanted to keep it backwards compatible.


Still wondering why it works in my system. set_autofilter_area is called in method create_worksheet of class zcl_excel_converter. This results in valid entries.
But of cause it would still fail, if no converter is used. I will check it.

darnoc312:
Converter's set_autofilter_area does not result in a valid entry immediately.
It is setting row_start, col_start and col_end but it leaves row_end initial. This is the issue @sandraros mentioned above.
This initial value triggers in autofilter's validate_area method the use of the value from worksheet's get_highest_row method
which finally results in a valid filter area.
Autofilter's validate_area method is necassary especially if the application program does not set any filter area at all.

darnoc312:
Hello,

please add a local field "lv_row_hidden TYPE abap_bool"  and try the following code to avoid too many calls of method is_row_hidden:
```abap
  LOOP AT lt_cell_data ASSIGNING <fs_sheet_content>.

*   inserted
    AT NEW cell_row.
      IF lo_autofilter IS NOT INITIAL.
        lv_row_hidden = lo_autofilter->is_row_hidden( iv_row = <fs_sheet_content>-cell_row ).
      ELSE.
        CLEAR lv_row_hidden.
      ENDIF.
    ENDAT.
*   end of inserted

* --- Add empty rows
    WHILE lv_row < <fs_sheet_content>-cell_row.
      CONCATENATE lv_string zcl_excel_writer_csv=>eol INTO lv_string.
      lv_row = lv_row + 1.
      lv_col = 1.
    ENDWHILE.

* --- Skip hidden rows
    IF lv_row_hidden = abap_true.               "changed
      lv_row = <fs_sheet_content>-cell_row + 1.
      lv_col = 1.
      CONTINUE.
    ENDIF.
```

oliver-huetkoeper:
Hi,

I added the suggested logic to improve the performance.

Best regards,
Oliver

oliver-huetkoeper:
Hi,

@sandraros what is the status on this? Is it ok now?

Thanks and best regards,
Oliver

sandraros (approved):
Thanks, LGTM

PR #1297, opened 2025-04-09
issue 1294 (cell comments)

rplantiko:
Enables rich text cell comments


gregorwolf:
@rplantiko you could correct this issue and push again to this PR.

sandraros:
There are many modifications in the PR, which make the PR difficult to approve, like Italian texts which we won't merge, improvements not related to the problem like default parameter values as constants, etc.

Is it possible to push only the code related to the fix?

To do this, I guess it's simpler to open a new PR.

I recommend using the "diff" button in abapGit, before staging, to make sure only the fixed code is pushed.

PR #1303, opened 2025-04-20
Remove italian in class components (SEOCOMPOTX)

### LINKED ISSUE #1299: abap2xlsx supported only in English but texts in Italian language exist
That's a follow up of discussion in #1297 and #1294 concerning Italian texts.

e.g. there is one text provided in both English and Italian languages (Italian text being the same as in English by the way):

https://github.com/abap2xlsx/abap2xlsx/blob/d9f557ea75c026e1a12b59b999765a83da69508b/src/zcl_excel.clas.xml#L30-L39

abap2xlsx objects are provided with main language set to English, hence abapGit installation requires to install in English language.

It's not said that abap2xlsx is provided with translations for one or more languages.

I would like to cleanup the translations to keep and maintain only English.

sandraros:
Fix #1299
Italian removed by running:
- `DELETE FROM SEOCOMPOTX WHERE ( CLSNAME LIKE 'ZCL_EXCEL%' OR CLSNAME LIKE 'ZCX_EXCEL%' OR CLSNAME LIKE 'ZIF_EXCEL%' ) AND LANGU = 'I'`
- Empty table buffer (all buffers in fact) to consider `SEOCOMPOTX` changes: `/$sync`


larshp (approved):
thanks

PR #1306, opened 2025-05-01
Allow XSTRING as input type for set_cell

bigld:
fixes 1091


larshp (approved):
thanks, looks good to me

PR #1315, opened 2025-06-05
Remove code duplication in check_rtf

rplantiko:
Method implementation of `zcl_excel->get_style_from_guid` is duplicated in `zcl_excel_worksheet->check_rtf`. 
This correction
- makes `zcl_excel->get_style_from_guid` public so that it can be called by `zcl_excel_worksheet `and other objects
- replaces the duplicate code in `zcl_excel_worksheet->check_rtf` by method call
- replaces multiple calls of `lo_style->font->get_structure( )` by a single call at the beginning
- repairs a bug in `zcl_excel_worksheet->check_rtf`, line 42: operator should be **<**, not **>**

### DIFF
--- a/src/zcl_excel.clas.abap
+++ b/src/zcl_excel.clas.abap
@@ -22,6 +22,11 @@ CLASS zcl_excel DEFINITION
         VALUE(ro_autofilter) TYPE REF TO zcl_excel_autofilter
       RAISING
         zcx_excel .
+    METHODS get_style_from_guid
+      IMPORTING
+        !ip_guid        TYPE zexcel_cell_style
+      RETURNING
+        VALUE(eo_style) TYPE REF TO zcl_excel_style .
     METHODS add_new_comment
       RETURNING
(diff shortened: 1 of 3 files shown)


sandraros (approved):
LGTM. I'm just adding a commit to move the position of the method `check_rtf` in the definitions, and to add unit tests for `check_rtf`.

darnoc312:
You could do even more: method check_rtf cannot deal with style guids representing static styles.

rplantiko:
> You could do even more: method check_rtf cannot deal with style guids representing static styles.

This PR is only a refactoring, doesn't add new functionality (although your proposal clearly makes sense). Would it be worth to make a new PR?

rplantiko:
> LGTM. I'm just adding a commit to move the position of the method `check_rtf` in the definitions, and to add unit tests for `check_rtf`.

Thanks for adding unit tests to safeguard the code paths.

sandraros:
> You could do even more: method check_rtf cannot deal with style guids representing static styles.

You can propose a different PR. For now, let's solve the current problem.

PR #1316, opened 2025-06-05
Wrap comment box parameters into structure

rplantiko:
The eight attributes `BOTTOM_OFFSET`, `BOTTOM_ROW`, `LEFT_COLUMN`, `LEFT_OFFSET`, `RIGHT_COLUMN`, `RIGHT_OFFSET`, `TOP_OFFSET `and `TOP_ROW ` of class `ZCL_EXCEL_COMMENT` belong together semantically. 

It is good style to put such attributes into  a common structure. This way, they can always be seen as a set belonging together, and can be passed around easier in the system, e.g. through method interfaces.  

I also added a structured constant containing their default values.

### DIFF
--- a/src/zcl_excel_comment.clas.abap
+++ b/src/zcl_excel_comment.clas.abap
@@ -5,8 +5,29 @@ CLASS zcl_excel_comment DEFINITION
 
   PUBLIC SECTION.
 
-    CONSTANTS default_right_column TYPE i VALUE 4.  "#EC NOTEXT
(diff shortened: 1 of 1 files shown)


sandraros (approved):
LGTM except I'm keeping the previous behavior for ip_right_column and ip_bottom_row to replace zero with the default value, although I don't like this old code, but what is important for me is to NOT make regressions in existing code using abap2xlsx.

rplantiko:
OK with your prioritizing backward compatibility over a probably better understandable and more consistent way of passing parameters.

PR #1317, opened 2025-06-13
Move copy code into copy constructor

rplantiko:
Small refactoring: The code in `zcl_excel_worksheet->get_comments` belongs into the class `zcl_excel_comments` itself, as this latter class is the expert on handling the comment collections. Also, the method `get_comments` creates a new comments object with a collection _copied_ from the existing one, so that changes of the new collection will not affect the original. This is classically the task of a copy constructor. So the code of `zcl_excel_worksheet->get_comments` moved into the constructor of `zcl_excel_comments`.

### DIFF
--- a/src/zcl_excel_comments.clas.abap
+++ b/src/zcl_excel_comments.clas.abap
@@ -12,7 +12,9 @@ CLASS zcl_excel_comments DEFINITION
       IMPORTING
         !ip_comment TYPE REF TO zcl_excel_comment .
     METHODS clear .
-    METHODS constructor .
+    METHODS constructor
+      IMPORTING
+        !io_from TYPE REF TO zcl_excel_comments OPTIONAL .
     METHODS get
       IMPORTING
         !ip_index         TYPE zexcel_active_worksheet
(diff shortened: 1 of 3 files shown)


darnoc312:
Does it make sense to make a copy at all ?
In my system i removed the copy for better performance several time ago.
Then the method contains the coding line 'r_comments = comments.' only.
Method get_drawings also doesn't make copies except you request both collectors.
Even in this case this is for reading or changing the elements in one loop,
but not for changing the copied collector itself by adding/removing elements.

rplantiko:
> Does it make sense to make a copy at all ? In my system i removed the copy for better performance several time ago. Then the method contains the coding line 'r_comments = comments.' only. Method get_drawings also doesn't make copies except you request both collectors. Even in this case this is for reading or changing the elements in one loop, but not for changing the copied collector itself by adding/removing elements.

I had the same question when I made that refactoring, but I wanted to preserve the functionality of the code as it is, and didn't put the decision of the author of that code into question. It must have been clear for him that the statement `r_comments = comments.` would have been easier, so I assumed that he had a purpose for his copy code. 

And, yes, it is only a "one-level" copy anyway (not a completely deep copy): the collection is copied, but the individual comment objects are (and were) still the original objects.  

The library itself may work with passing the reference. But maybe there is code of some users which will be broken? It would be strange, however, if any user wants to change the collection without changing the original, but maybe I am just lacking imagination. 

But okay: if the general opinion is to work directly with the reference here, we could use this PR to remove the copy process completely. It would simplify the code and make it faster - at the price of breaking a backward compatibility, but at a point where it doesn't have much use anyway. If somebody consciously wants to make a copy of the collection, he could use the copy constructor (so we could leave this part of the code, and only replace the `get_comments` implementation by the simple statement `r_comments = comments.`

rplantiko:
I now changed the `zcl_excel_worksheet->get_comments` implementation into `r_comments = comments.`
This is a functional change, but there is no conceivable benefit of copying the collection. 
Are there objections against the merge?

sandraros:
I agree with all comments here that the existing code is not very good (like in many other places), but why risking regressions in programs using abap2xlsx? I would keep `create object r_comments` each time `get_comments` is invoked. If you wish, you may add an optional parameter to not instantiate. The default value should do `create object r_comments`.

rplantiko:
To keep you both happy, I made 

- copying the collection the default behaviour of  `get_comments( )`
- allowed retrieving the `comments` object directly, when the new import parameter `iv_copy_collection` is set to `abap_false`.

Learned: in this project, backward compatibility is valued **very** high.

sandraros (approved):
LGTM

PR #1323, opened 2025-07-05
Rework for class zcl_excel_comment

### LINKED ISSUE #1322: Rework for class zcl_excel_comment after PR #1219 and PR #1316
After the changes of PR #1219 and PR #1316 the class zcl_excel_comment needs a rework.

Remember:

1) If visibility is set to hidden (this is the ABAP2XLSX standard, see zcl_excel_writer_2007 current line 3278)
the comment's position is always adjusted right (col+1) above (line-1) to its referenced cell position.
That means that the user can affect only width and height of the comment's box.
Currently the user can set width and height only by differences between right and left column or bottom and top row respectively.
This is weird and unfortunately.

2) If visibility in excel is set to on the comment box will be placed at the specified coordinates starting at 0.
Do you really think that your snapshot at the end of #1221 shows the result to achieve ?
In my opinion it should look like this:

Image

-The current standard values for left_column (2), right_column (4) , top_row (11) and bottom_row (15) don't
contain any intention to be preserved for any compatibility reason.
These are the values for the first comment ever set at B13 ("This is how it begins to be debug time")
while developing program ZDEMO_EXCEL_COMMENTS and corresponds for this cell to my description above
(adjusted to its referenced cell position).
And then they were never changed again for other cells, because it doesn't matter if visibility is set to hidden.
But these values makes no sense for any other cell coordinates if not explicitly requested.

-The method set_text is not reusable currently.
If i use the box values in one call and then call set_text again for the same object these values are all reset to default.
But for the reading part of #1294 i need a call to set the box data first and then a second call to set the text and RTF Tab.
Currently this is not possible.

darnoc312:
fix #1322 

1) I want to implement the following call in program ZDEMO_EXCEL_COMMENTS replacing the old one
```abap
  lo_comment->set_text( ip_text          = 'A comment at any position and any size'
                        ip_ref           = 'C18'
                        ip_topleft       = 'E20'
                        ip_width         = 2
                        ip_height        = 3
                        ip_left_offset   = 0
                        ip_top_offset    = 0
                        ip_right_offset  = 0
                        ip_bottom_offset = 0 ).
```
ip_topleft is optional. Other calls can pass ip_width and ip_height only.

ip_topleft ist also passed as reference like ip_ref.
It is weird for the user that the ref starts at line 1/column A(=1) and the 4 values for ip_left_column etc. start from 0.

2) Anything i marked with "obsolet" might be removed or might be preserved for compatibility reasons.
The 4 parameters ip_left_column, ip_right_column, ip_top_row and ip_bottom_row
The method set_box was unfortunately defined as public by Rüdiger Plantiko. 
Two public definitions at the beginning of the class.
Please tell me what you like to delete.

3) The changes here do not overlap with #1321. So it is mergable cleanly.

### DIFF
--- a/src/zcl_excel_comment.clas.abap
+++ b/src/zcl_excel_comment.clas.abap
@@ -5,6 +5,7 @@ CLASS zcl_excel_comment DEFINITION
 
   PUBLIC SECTION.
 
+    "obsolet, do not use
     TYPES:
       BEGIN OF ty_box,
         left_column   TYPE i,
@@ -28,6 +29,7 @@ CLASS zcl_excel_comment DEFINITION
         bottom_row    TYPE i VALUE 15,
         bottom_offset TYPE i VALUE 9,
       END OF gc_default_box .
+    "end of obsolet
 
     METHODS constructor .
     METHODS get_bottom_offset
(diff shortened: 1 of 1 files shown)


V-j11:
> fix #1322 
> 
> 
> 
> 1) I want to implement the following call in program ZDEMO_EXCEL_COMMENTS replacing the old one
> 
> ```abap
> 
>   lo_comment->set_text( ip_text          = 'A comment at any position and any size'
> 
>                         ip_ref           = 'C18'
> 
>                         ip_topleft       = 'E20'
> 
>                         ip_width         = 2
> 
>                         ip_height        = 3
> 
>                         ip_left_offset   = 0
> 
>                         ip_top_offset    = 0
> 
>                         ip_right_offset  = 0
> 
>                         ip_bottom_offset = 0 ).
> 
> ```
> 
> ip_topleft is optional. Other calls can pass ip_width and ip_height only.
> 
> 
> 
> ip_topleft ist also passed as reference like ip_ref.
> 
> It is weird for the user if the ref starts at line 1 and column A(=1) and the 4 values for ip_left_column etc. start from 0.
> 
> 
> 
> 2) Anything i marked with "obsolet" might be removed or might be preserved for compatibility reasons.
> 
> The 4 parameters ip_left_column, ip_right_column, ip_top_row and ip_bottom_row
> 
> The method set_box was unfortunately defined as public by Rüdiger Plantiko. 
> 
> Two public definitions at the beginning of the class.
> 
> Please tell me what you like to delete.
> 
> 
> 
> 3) The changes here do not overlap with #1321. So it is mergable cleanly.
> 
>

PR #1298, opened 2025-04-09
RTF cell comments

rplantiko:
Introduces rich text cell comments

PR #1327, opened 2025-07-06
Constants and Data reorganized to prepare #1298

sandraros:
Refactor the method, to make it close to the actual changes done by #1298: all CONSTANTS replaced by literals and DATA lines sorted according to #1298.


darnoc312:
But you don't really consider to merge #1298 ?

sandraros:
> But you don't really consider to merge #1298 ?

If you have some comments on #1298, please comment there. Thank you.

darnoc312:
What is the advantage of doing this ?
To have an overview about the correct spelling (including lower/upper case) of the keywords it is better to have them in a CONSTANTS Block.

PR #757, opened 2021-05-28
Fix of bug #756

sandraros:
This is a solution to complement PR #743

Vietnamese/Emoji should now be rendered in systems before ABAP 7.52

PR #1349, opened 2025-10-15
Implementing the part of SAP Note 2922674 concerning the reader

### LINKED ISSUE #1347: older 7.31 system renders unicode icons differently
On 731 system, logon EN to avoid spurious differences due to language, the checker reports the following differences in demos:

- 37: Emojis are not rendered (backend too old)

Image

- 38: icons are rendered differently, especially red one

Image

- 15_01: consequence of 37

The same check on a more modern system passes just fine.

@sandraros creating in main project as you suggested, we may move it to demos later if it turns out to be specific to those test programs. Also, is it possible that the icons have changed in SAP over time?

darnoc312:
fix #1347

SAP Note 2922674 (concerning SAP_BASIS release <= 7.51) was implemented in the writer (method render_xml_document, please refer to PR #757) but not in the reader so far

I got the same issue on a 740 unicode system. After this change file 15_01_HelloWorldFromReader.xlsx looks nicely again.

### DIFF
--- a/src/zcl_excel_reader_2007.clas.abap
+++ b/src/zcl_excel_reader_2007.clas.abap
@@ -652,6 +652,12 @@ CLASS zcl_excel_reader_2007 IMPLEMENTATION.
 
   METHOD get_ixml_from_zip_archive.
 
+* The corresponding part of SAP note 2922674:
+* A workaround for the replacement of characters of the supplementary plane in SAP_BASIS 7.51 or lower is to convert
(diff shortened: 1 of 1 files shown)


AndreaBorgia-Abo (approved):
This DOES solve the issue, running ZDEMO_EXCEL_CHECKER:
* before PR: fails 10, 22, 37, 38, 47, 49, 15_01
* after PR: fails 10, 22, 38, 47, 49

Thanks for the PR, nice work!

PR #1356, opened 2026-01-20
Performance improvement in writer when having many styles

### LINKED ISSUE #1355: Performance: inefficient table type in ZCL_EXCEL_WRITER_2007
Hi,

When creating big files with many styles there might be a massive performance degradation.

Traces show that the root cause is attribute zcl_excel_writer_2007->styles_mapping which is a standard table, that is usually read "WITH KEY guid = ...", however, as the table has no hash key, this results in a full table scan, which can be very costly if the table is big (many style guids) and read a lot of times (many cells) and thus shows parablic performane degradation.

The table should be switched to hashed table of technically possible which will speed up the read operations and improve performance. Other similar tables should also be evaluated.

It should be noted that the hight number of style guids is also an 'application problem', that is the application using abap2xlsx probably should 'pool' styles, nevertheless the technical optimization if the table - if feasible - makes sense.

PR will follow.

oliver at evosight

oliversviszt:
Fixes: #1355 

**Problem:** Poor performance when writing files with many cells using many different styles.
**Cause:** Standard table for style guid->index mapping with full table scan, runtim O = n^2
**Fix:** Change table to hashed table
**Testing:** unit tests, demo output tests, syntax check. Performance increase for file with 52.000 cells and styles: runtime down from 15 to 2 seconds.

Note: DDIC table type zexcel_t_styles_mapping becomes obsolete but was not deleted (and neither adapted) because it might be used in userland code.


larshp:
keeping the TTYP will potentially cause runtime errors, and removing it will give static errors? So suggest removing it

oliversviszt:
Hi @larshp 
Thanks for the feedback, I have now deleted the TTYP. Abaplint check now successful.

Oliver

darnoc312:
1) In my opinion you should remove structure ZEXCEL_S_STYLES_MAPPING, too.
Please insert the following TYPES definition after the mty_column_formulas_used type definition:
```abap
  types:
    BEGIN OF mty_style_mapping,
      guid  TYPE zexcel_cell_style,
      style TYPE i,
    END OF mty_style_mapping .
  types:
    mty_styles_mapping TYPE HASHED TABLE OF mty_style_mapping WITH UNIQUE KEY guid .
```
and replace the references by "LIKE LINE OF me->styles_mapping"


2) Please try the following program (in particular the huge file mode)
```abap
REPORT zdemo_excel_style_mapping.

DATA:
  lo_excel        TYPE REF TO zcl_excel,
  lo_worksheet    TYPE REF TO zcl_excel_worksheet,
  lo_style        TYPE REF TO zcl_excel_style,
  lo_column       TYPE REF TO zcl_excel_column,
  lv_column_alpha TYPE zexcel_cell_column_alpha.

CONSTANTS:
  gc_save_file_name TYPE string VALUE 'DemoStyleMapping.xlsx'.

PARAMETERS: p_normal RADIOBUTTON GROUP rb DEFAULT 'X',
            p_huge   RADIOBUTTON GROUP rb.


START-OF-SELECTION.

  CREATE OBJECT lo_excel.
  lo_worksheet = lo_excel->get_active_worksheet( ).

  DO 5 TIMES.
    lv_column_alpha = zcl_excel_common=>convert_column2alpha( sy-index ).
    lo_style = lo_excel->add_new_style( ).
    lo_style->font->color-rgb = zcl_excel_style_color=>c_red.
    lo_column = lo_worksheet->get_column( lv_column_alpha ).
    lo_column->set_column_style_by_guid( lo_style->get_guid( ) ).
    lo_worksheet->set_cell( ip_column = lv_column_alpha ip_row = 2 ip_value = lv_column_alpha ).
  ENDDO.

  CASE abap_true.
    WHEN p_normal.
*** Create output

    WHEN p_huge.
*** Create output
      lcl_output=>output( cl_excel = lo_excel iv_writerclass_name = 'ZCL_EXCEL_WRITER_HUGE_FILE' ).

  ENDCASE.
```

Writer_huge_file is the only caller of ZCL_EXCEL's method GET_STYLE_INDEX_IN_STYLES (two times). But this method is useless.
It should be at least marked as obsolete or even removed. It does not achieve the result of an access to the styles mapping table.
If more than one A2X style objects with the same content are merged to one resulting excel style number (by WRITER's method CREATE_XL_STYLES)
this method fails.


Please replace in line 491 and 532 of ZCL_EXCEL_WRITER_HUGE_FILE:
from:
```abap
      <col>-style = me->excel->get_style_index_in_styles( lv_style_guid ) - 1.
```

to:
```abap
      <col>-style = -1.
      IF lv_style_guid IS NOT INITIAL.
        READ TABLE styles_mapping ASSIGNING <style_mapping> WITH TABLE KEY guid = lv_style_guid.
        IF sy-subrc = 0.
          <col>-style = <style_mapping>-style.
        ENDIF.
      ENDIF.
```

and add an appropiate field-symbol to the other field-symbols starting at line 308:
```abap
    <style_mapping> LIKE LINE OF me->styles_mapping,
```

oliversviszt:
Thanks for your feedback.

1. Done. (Did not use LIKE LINE OF but direct reference to the structure type.)

2. I understand the problem and have implemented the suggested changes. (With INTO instead of ASSIGNING as on modern systems this brings no performance benefit.) 

Unit test and demo files are fine, however, when comparing your test case from zdemo_excel_style_mapping the color is not applied properly in Excel in the file generate by Huge-Writer.
This did not work before the changes either in my opinion, before the adaptations there were some weird formats, now it is simply that the style is not applied and the text remains black.

styles.xml seems identical, besides sheet1.xml other differences also seem marginal.
Here is a comparison:
image

So it seems Huge-Writes writes no cell (<c>) level styles (<c s=...>) where the classic writer does. No matter what I have tried, <col styles=...> is only applied as a default formatting when entering values in Excel into cells which were empty so far, but it is not inherited to existing cells. In other words, it is a default, not a fallback.

I don't think this different should be part of our current issue. If at all, we should create a new issue and tackle it there.

oliversviszt:
Hi @larshp, seems that the abaplint checks are stuck since 11 hours. Is there any specific piece in my code which causes this, or a service issue?

larshp:
lots of problems with github service yesterday, try pushing another commit, just do a dummy change and revert it with another commit

darnoc312:
Yes, this is another bug in writer_huge_file. method get_cells doesn't look for any excel-, worksheet- or column-default style (like writer_2007 does) but only for the cell style which here is empty.

V-j11 (approved):

PR #1358, opened 2026-02-16
Non-Unicode fixes

michaelsargent:
These fixes check for a non-Unicode system to make the demo programs work and avoid code page conversion dumps.
Tested on a 740 ABAP system with non-Unicode.

### DIFF
--- a/src/zcl_excel_common.clas.abap
+++ b/src/zcl_excel_common.clas.abap
@@ -218,7 +218,7 @@ CLASS zcl_excel_common DEFINITION
         zcx_excel .
     CLASS-METHODS is_unicode_system
       RETURNING
-        VALUE(rv_is_unicode) TYPE abap_bool .		
+        VALUE(rv_is_unicode) TYPE abap_bool .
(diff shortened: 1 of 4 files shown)


src/zcl_excel_common.clas.abap
```diff
@@ -1042,6 +1045,29 @@ CLASS zcl_excel_common IMPLEMENTATION.
   ENDMETHOD.
 
 
+  METHOD is_unicode_system.
+
+    " Cache the result as static variable for performance
+    STATICS: sv_cached     TYPE abap_bool,
+             sv_is_unicode TYPE abap_bool.
```
larshp: is it possible to use CLASS-DATA instead of statics?

michaelsargent:
I have tested demo programs successfully on on-premise S4 Hana 2023 system.

src/zcl_excel_common.clas.abap
michaelsargent: I removed the statics and used a simpler approach.

